PR #9, opened 2015-01-23
Added local stream wrapper for local development

joehoyle:


inc/class-s3-uploads-local-stream-wrapper.php
```diff
@@ -0,0 +1,499 @@
...
+		if (!isset($uri)) {
+			$uri = $this->uri;
+		}
+		$path = $this->getDirectoryPath() . '/' . $this->getTarget($uri);
+		$realpath = $path;
+		
+		$directory = realpath($this->getDirectoryPath());
+
+		if (!$realpath || !$directory || strpos($realpath, $directory) !== 0) {
+			return FALSE;
```
rmccue: `false` should be lowercase :)

inc/class-s3-uploads-local-stream-wrapper.php
```diff
@@ -0,0 +1,499 @@
...
+	 */
+	protected $uri;
+
+	/**
+	 * Gets the path that the wrapper is responsible for.
+	 *
+	 * @return string
+	 *   String specifying the path.
+	 */
+	static function getDirectoryPath() {
```
rmccue: These should have `public` specifiers to be explicit

inc/class-s3-uploads.php
```diff
@@ -30,11 +30,14 @@ public function __construct( $bucket, $key, $secret, $bucket_hostname = null ) {
...
-		
-
+		if ( defined( 'S3_UPLOADS_USE_LOCAL' ) && S3_UPLOADS_USE_LOCAL ) {
+			require_once dirname( __FILE__ ) . '/class-s3-uploads-local-stream-wrapper.php';
+			stream_wrapper_register( 's3', 'S3_Uploads_Local_Stream_Wrapper', STREAM_IS_URL );
+		} else {
+			$this->s3()->registerStreamWrapper();
+			stream_context_set_option( stream_context_get_default(), 's3', 'seekable', true );
+			stream_context_set_option( stream_context_get_default(), 's3', 'ACL', Aws\S3\Enum\CannedAcl::PUBLIC_READ );
+		}
```
rmccue: We should set these context options always, and only do the minimal amount of difference.

inc/class-s3-uploads.php
```diff
@@ -45,8 +48,14 @@ public function filter_upload_dir( $dirs ) {
...
 		if ( ! defined( 'S3_UPLOADS_DISABLE_REPLACE_UPLOAD_URL' ) || ! S3_UPLOADS_DISABLE_REPLACE_UPLOAD_URL ) {
-			$dirs['url'] = str_replace( WP_CONTENT_URL, $this->get_s3_url(), $dirs['url'] );
-			$dirs['baseurl'] = str_replace( WP_CONTENT_URL, $this->get_s3_url(), $dirs['baseurl'] );
+
+			if ( defined( 'S3_UPLOADS_USE_LOCAL' ) && S3_UPLOADS_USE_LOCAL ) {
+				$dirs['url']     =  str_replace( WP_CONTENT_URL, $dirs['baseurl'] . '/s3/' . str_replace( $this->bucket_hostname, strtok( $this->bucket, '/' ), $this->get_s3_url() ), $dirs['url'] );
+				$dirs['baseurl'] = str_replace( WP_CONTENT_URL, $dirs['baseurl'] . '/s3/' . str_replace( $this->bucket_hostname, strtok( $this->bucket, '/' ), $this->get_s3_url() ), $dirs['baseurl'] );
+			} else {
+				$dirs['url']     = str_replace( WP_CONTENT_URL, $this->get_s3_url(), $dirs['url'] );
+				$dirs['baseurl'] = str_replace( WP_CONTENT_URL, $this->get_s3_url(), $dirs['baseurl'] );
```
rmccue: Can we make these equal?

inc/class-s3-uploads.php
joehoyle: Fixed in 042e67f

inc/class-s3-uploads-local-stream-wrapper.php
joehoyle: As this is a Drupal file, I'm going to leave formatting as is.

inc/class-s3-uploads.php
joehoyle: The complication here is that we are putting the `s3` local dir _in_ the uploads dir, this means the usual str_replace messes up as it always has the upload dir in it, even the new path. Of course we could store it in `root/s3/` but I think that's more error prone with permissions etc.

joehoyle:
Merging as is, though @rmccue feel free to response re the "Can we make these equal?" comment.

inc/class-s3-uploads.php
rmccue: > The complication here is that we are putting the `s3` local dir in the uploads dir, this means the usual str_replace messes up as it always has the upload dir in it, even the new path. Of course we could store it in `root/s3/` but I think that's more error prone with permissions etc.

Ah, gotcha, makes sense. I'd prefer to keep them as close as possible, but makes sense if we can't. :)

(That said, is there a reason it's here rather than in `$this->get_s3_url()`?)

rmccue:
> Merging as is, though @rmccue feel free to response re the "Can we make these equal?" comment.

:+1:

joehoyle:
Had a couple more fixes in 26803ac

PR #16, opened 2015-06-05
Adding a WP-CLI Verify command

benmay:
We added a verify command to easily diagnose when incorrect API keys were used. It does an S3 upload then deletes it testing both of those rules. 

May not be so useful if you use the generated policy, but we used our own existing policy and realised the way the AWS API works on PHP is a bit more demanding than S3CMD and it wasn't obvious that the keys were wrong.

Hopefully, this is a handy / useful feature for others.


inc/class-s3-uploads-wp-cli-command.php
```diff
@@ -255,19 +309,19 @@ public function upload_directory( $args, $args_assoc ) {
 		require_once dirname( __FILE__ ) . '/class-s3-uploads-changed-files-iterator.php';
 
 		try {
-			$s3->uploadDirectory(
```
rmccue: In general, avoid unrelated whitespace changes, since they can cause merge conflicts.

inc/class-s3-uploads-wp-cli-command.php
```diff
@@ -3,6 +3,60 @@
 class S3_Uploads_WP_CLI_Command extends WP_CLI_Command {
 
 	/**
+     * Verifies the API keys entered will work for writing and deleting from S3.
```
rmccue: Looks like your indentation might be off here?

inc/class-s3-uploads-wp-cli-command.php
benmay: Yep. Worked out with @zacscott what his editor was doing.

inc/class-s3-uploads-wp-cli-command.php
```diff
@@ -3,6 +3,60 @@
...
+		$upload_dir = wp_upload_dir();
+
+		// The upload file location on the local filesystem
+		$s3_path = $upload_dir['basedir'] . '/' . mt_rand() . '.jpg';
+
+		// Attempt to copy the file to S3
+		WP_CLI::print_value( 'Attempting to upload file '. $s3_path );
+		
+		// Set Error Handler so we can catch it if it fails 
+		set_error_handler( array( $this, 'copy_fail' ) );
```
joehoyle: You can catch it failing by `copy` returning false. I'd rather not set a global error handler if possible, it makes things more complex.

inc/class-s3-uploads-wp-cli-command.php
```diff
@@ -3,6 +3,60 @@
 class S3_Uploads_WP_CLI_Command extends WP_CLI_Command {
 
 	/**
+	 * Verifies the API keys entered will work for writing and deleting from S3.
+	 *
+	 * @subcommand verify
+	 */
+    public function verify_api_keys() {
+
+		S3_Uploads::get_instance(); // Boot
```
joehoyle: This should already be hooked on the normal plugin includes, or is there some other reason why this is needed?

inc/class-s3-uploads-wp-cli-command.php
```diff
@@ -3,6 +3,60 @@
...
+
+		// Copies canolda from the test dir, upto S3
+		$copy = copy(
+			dirname( dirname(__FILE__) ) . '/tests/data/canola.jpg',
+			$s3_path
+		);
+
+		restore_error_handler();
+
+		WP_CLI::print_value( 'File uploaded to S3 successfully' );
```
joehoyle: `WP_CLI::success(` is probably better here.

inc/class-s3-uploads-wp-cli-command.php
```diff
@@ -3,6 +3,60 @@
...
+
+		// Delete off S3
+		WP_CLI::print_value( 'Attempting to delete file '. $s3_path );
```
joehoyle: spacing around `.` needed on the left (several other examples here too)

inc/class-s3-uploads-wp-cli-command.php
```diff
@@ -3,6 +3,60 @@
...
+			return;
+		}
+
+		WP_CLI::print_value( 'File deleted from S3 successfully' );
+
+		WP_CLI::success( 'Looks like your configuration is correct.' );
+
+    }
+
+    /**
```
joehoyle: This is using spaces, not tabs - though we might want to remove, per the comment above.

joehoyle:
Thanks for the PR @benmay, much appreciated! A few small notes, but this is looking good.

inc/class-s3-uploads-wp-cli-command.php
zacscott: We did try using the copy() result first, however it did not work. It would always return true, even if it failed. The only thing that we could do was grab the warning it spits out.

joehoyle: @zacscott ah in that case, we should fix why copy is returning true, because it shouldn't do. Maybe we could get a failing unit test for this case.

joehoyle: @zacscott I fixed the bug where `copy` would return `false` on fail - so we this should be ok to update now if you merge `master` into our branch.

zacscott: Excellent, thanks for the fix Joe. I was going to have a look but you saved me the trouble :-)  I will update now

joehoyle:
@benmay thanks!

aaronk6:
@benmay Thanks for adding this!

I’m just wondering if it’s possible to have wp-cli return an error code when this fails. I’m using this in a deployment script and rely on the exit code to learn if it verified successfully or not. Currently, it returns always 0, even when it has failed:

```
$ wp s3-uploads verify
Error: The required constant S3_UPLOADS_BUCKET is not defined.
Error: The required constant S3_UPLOADS_KEY is not defined.
Error: The required constant S3_UPLOADS_SECRET is not defined.
$ echo $?
0
```

This should return 1 (or anything other than 0) instead.

Is this something you can fix or is this a general issue with wp-cli?

PR #17, opened 2015-06-05
Add filter for modifying S3 Object Paramaters

benmay:
We wanted to set defaults on Expires and cache controls, so we added a filter so you could modify the array passed to S3 via an external plugin / drop in.

However, since Expires & Cache-Control are probably quite popular, we added support for constants to be added to wp-config with the API keys to control this without requiring an additional drop-in to configure.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ wp s3-uploads upload-directory <from> <to> [--sync] [--dry-run]
 
 Passing `--sync` will only upload files that are newer in `<from>` or that don't exist on S3 already. Use `--dry-run` to test.
 
-There is also an all purpose `cp` command for arbitraty copying to and from S3.
+There is also an all purpose `cp` command for arbitrary copying to and from S3.
 
 ```
 wp s3-uploads cp <from> <to>
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ wp s3-uploads create-iam-user --admin-key=<key> --admin-secret=<secret>
 
 This will provide you with a new Access Key and Secret Key which you can configure S3-Uploads with. Paste the values in the `wp-config.php`. Once you have migrated your media to S3 with any of the below methods, you'll want to enable S3 Uploads: `wp s3-uploads enable`.
 
-If you want to create your IAM user yourself, or attach the neccessary permissions to an existing user, you can output the policy via `wp s3-uploads generate-iam-policy` 
+If you want to create your IAM user yourself, or attach the neccessary permissions to an existing user, you can output the policy via `wp s3-uploads generate-iam-policy`
 
 Migrating your Media to S3
 ==========
@@ -61,10 +61,30 @@ wp s3-uploads upload-directory <from> <to> [--sync] [--dry-run]
 
 Passing `--sync` will only upload files that are newer in `<from>` or that don't exist on S3 already. Use `--dry-run` to test.
 
-There is also an all purpose `cp` command for arbitraty copying to and from S3. 
+There is also an all purpose `cp` command for arbitraty copying to and from S3.
 
 ```
 wp s3-uploads cp <from> <to>
 ```
 
 Note: as either `<from>` or `<to>` can be S3 or local locations, you must speficy the full S3 location via `s3://mybucket/mydirectory` for example `cp ./test.txt s3://mybucket/test.txt`.
+
+Cache Control
+==========
+
+You can define the default HTTP `Cache-Control` header for uploaded media using the
+following constant:
+
+```PHP
+define( 'S3_UPLOADS_CACHE_CONTROL', 30 * 24 * 60 * 60 );
+	// will expire in 30 days time
+```
+
+You can also configure the `Expires` header using the `S3_UPLOADS_EXPIRES` constant
+For instance if you wanted to set an asset to effectively not expire, you could
+set the Expires header way off in the future.  For example:
+
+```PHP
+define( 'S3_UPLOADS_EXPIRES', gmdate( 'D, d M Y H:i:s', time() + (10 * 365 * 24 * 60 * 60) ) .' GMT' );
+	// will expire in 10 years time
+```
--- a/inc/class-s3-uploads-stream-wrapper.php
+++ b/inc/class-s3-uploads-stream-wrapper.php
@@ -20,6 +20,22 @@ class S3_Uploads_Stream_Wrapper extends Aws\S3\StreamWrapper {
 	// Override
 	public function stream_flush() {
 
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -61,10 +61,30 @@ wp s3-uploads upload-directory <from> <to> [--sync] [--dry-run]
 
 Passing `--sync` will only upload files that are newer in `<from>` or that don't exist on S3 already. Use `--dry-run` to test.
 
-There is also an all purpose `cp` command for arbitraty copying to and from S3. 
+There is also an all purpose `cp` command for arbitraty copying to and from S3.
```
rmccue: lol "arbitraty", nice one @joehoyle

inc/class-s3-uploads-stream-wrapper.php
```diff
@@ -17,7 +17,34 @@ public static function register( Aws\S3\S3Client $client)
...
+			} else {
+				$this->params[ 'CacheControl' ] = S3_UPLOADS_CACHE_CONTROL;
+			}
+		}
+
+		// Theses are the parameters passed to S3Client::putObject()
+		// http://docs.aws.amazon.com/aws-sdk-php/latest/class-Aws.S3.S3Client.html#_putObject
+		$this->params = apply_filters( 's3_uploads_putObject_params', $this->params );
+
+		parent::stream_flush();
```
rmccue: Can we also `return` this, just to be sure?

inc/class-s3-uploads-stream-wrapper.php
```diff
@@ -17,7 +17,34 @@ public static function register( Aws\S3\S3Client $client)
...
+
+	}
+
 	public function stream_metadata( $path, $option, $value ) {
 		// not implemented
 	}
-}
\ No newline at end of file
+}
```
rmccue: :heart_eyes:

inc/class-s3-uploads-stream-wrapper.php
```diff
@@ -17,7 +17,34 @@ public static function register( Aws\S3\S3Client $client)
...
 	}
 
+	// Override
+	public function stream_flush() {
+
+		// Add some HTTP headers which can be defined in wp-config.php
+
+		// Expires:
+		if ( defined( 'S3_UPLOADS_EXPIRES' ) ) {
+			$this->params[ 'Expires' ] = S3_UPLOADS_EXPIRES;
```
rmccue: This will create an array implicitly and throw a strict standards warning (IIRC). Can we add `protected $params = array()` as a property to fix this please?

README.md
benmay: Will pop in a fix while I have it open.

inc/class-s3-uploads-stream-wrapper.php
benmay: Sure thing.

inc/class-s3-uploads-stream-wrapper.php
benmay: Yea..

inc/class-s3-uploads-stream-wrapper.php
benmay: This class extends the parent which already defines `$params` 

If I `var_dump` at the beginning of our `stream_flush()`, the `$params` var is setup and contains 

```
array(3) {
  ["Bucket"]=>
  string(7) "s3-bucket-name"
  ["Key"]=>
  string(26) "uploads/2015/06/stock5.jpg"
  ["ACL"]=>
  string(11) "public-read"
}
```

So I don't think we'd need to add that in, right?

joehoyle: I think we need a better name for the constant. On first glance, I thought this was about the object lifecycle on s3 - rather than the browser cache expires header (or maybe this is for object expires lifcycle? :confused: )

inc/class-s3-uploads-stream-wrapper.php
```diff
@@ -17,7 +17,35 @@ public static function register( Aws\S3\S3Client $client)
...
+		}
+
+		// Cache-Control:
+		if ( defined( 'S3_UPLOADS_CACHE_CONTROL' ) ) {
```
joehoyle: Ideally these constant should be passed into the constructor, so the caller controls whether to use the constants for not, this way the constants are decoupled form the class. It's possible someone would want to use their own instance of this class without the global constants setting these options.

See https://github.com/humanmade/S3-Uploads/blob/master/inc/class-s3-uploads.php#L20 for a similar example - the Shared Instance uses the wp-config constants, but it's still possible to initiate a new instance with your own details if you want. Though we don't have  a shared instance here - a similar logic can be applied.

inc/class-s3-uploads-stream-wrapper.php
```diff
@@ -17,7 +17,35 @@ public static function register( Aws\S3\S3Client $client)
...
+			if ( is_numeric( S3_UPLOADS_CACHE_CONTROL ) ) {
+				$this->params[ 'CacheControl' ] = 'max-age='. S3_UPLOADS_CACHE_CONTROL;
+			} else {
+				$this->params[ 'CacheControl' ] = S3_UPLOADS_CACHE_CONTROL;
+			}
+		}
+
+		// Theses are the parameters passed to S3Client::putObject()
```
joehoyle: I haven't looked into it, but is there some more Core Official way of documenting these hooks?

joehoyle:
Thanks for the PR, couple questions and notes, let me know what you think, thanks!

inc/class-s3-uploads-stream-wrapper.php
rmccue:
```php
/**
 * Filter the parameters passed to S3
 *
 * ...
 * @param array $params Parameters to filter
 * @param ... (for any other params passed in)
 */
```

Also, we should avoid mixing case here.

zacscott: Done

inc/class-s3-uploads-stream-wrapper.php
zacscott: I have renamed them to include HTTP.  i.e. S3_UPLOADS_HTTP_EXPIRES and S3_UPLOADS_HTTP_CACHE_CONTROL

inc/class-s3-uploads-stream-wrapper.php
zacscott: Don't think we can do this.  The class is registered using stream_wrapper_register() in S3_Uploads_Stream_Wrapper::register()

joehoyle:
@benmay thanks!

PR #19, opened 2015-06-12
Add composer support

spacedmonkey:
This is a simple enough change, adding this file means that you can use this repo to load via composer. 

There are only a couple of things to consider with this. 

1.  Should the type be "wordpress-plugin"? Is this code really designed to be an mu-plugin or is it meant to be a plugin?
2. You can use composer to get dependancies. This means you wouldn't have to hardcode the aws-sdk into the repo and using composer to grab the latest stable version. It already seems to support composer as noted [here](https://github.com/aws/aws-sdk-php/blob/master/composer.json). But it also brings it [own dependancies](https://github.com/aws/aws-sdk-php/blob/master/composer.json#L20), so might be a pain. This might be one for future development. I know that @danielhomer was looking into this.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,0 +1,24 @@
+{
+	"name": "humanmade/s3-uploads",
+	"description": "WordPress plugin to store uploads on S3",
+	"homepage": "https://github.com/humanmade/S3-Uploads",
+	"keywords": [
+		"wordpress"
+	],
+	"license": "GPL-2.0+",
+	"authors": [
+		{
+			"name":"Human Made Limited",
+			"email":"support@humanmade.co.uk",
+			"homepage":"http://hmn.md/"
+		}
+	],
+	"support"    : {
+			"issues": "https://github.com/humanmade/s3-uploads/issues",
+			"source": "https://github.com/humanmade/s3-uploads"
+	},
+	"type": "wordpress-plugin",
+	"require": {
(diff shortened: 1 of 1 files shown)


danielhomer:
The only reason I abandoned loading the AWS SDK via composer was because it needed hacking (I think it was something to do with the seek method in Guzzle). I also noticed a couple of hacks by Joe in the bundled version.

joehoyle:
I'd rather not have composer as a dependency, hence not using Composer to bring in the AWS SDK. As @danielhomer noted we have a couple hacks to the SDK too to fix some edge case issues with the streamwrapper, however it may be possible to remove these now as we are using a subclass of the streamwrapper, either way - we wouldn't switch to managing dependancies with composer.

> Should the type be "wordpress-plugin"? Is this code really designed to be an mu-plugin or is it meant to be a plugin?

It's designed to be a Plugin - though typically I'd suggest using it as a mu-plugin, I don't really distinguish between the two.

I think due to the secrets for S3 access in TravisCI, the tests fail, as for security reasons it doesn't include the secrets in 3rd party pull-requests. I think I have to pull down you branch and open a PR with that.

joehoyle:
Appears I just have to pull the commits into our `origin` for it to pass the tests, which make sense

joehoyle:
@spacedmonkey thanks!

PR #23, opened 2015-06-13
Move the aws-sdk modifications to our own class

joehoyle:
This moves the hacks we did have to the sdk into our own subclass, which makes
upgrading the SDK a lot simpler.


joehoyle:
Need to also pull put the hacks in committed in https://github.com/humanmade/S3-Uploads/commit/a9db1096ca5a042386073851c96061102265b309

joehoyle:
Merging this for now - created #26 to cover the last one.

Issue #15, opened 2015-06-03
Enabling the plugin via CLI 


benmay:
The docs indicate you need to run `wp s3-uploads enable` to enable the plugin, however when run, CLI returns and error `Error: 's3-uploads enable' is not a registered wp command. See 'wp help'.` and uploading didn't work or return any errors with all the usual debugging on.

benmay:
(The plugin works - but maybe that ref in the doc can come out)

rmccue:
I think it's meant to be `wp plugin enable s3-uploads` ?

rmccue:
Or not, ¯\_(ツ)_/¯

joehoyle:
Good point - I think in the past I had planned for it to be disabled until one had uploaded all existing uploads to s3, and then you could flip the "enabled" switch - but that never came to fruition, however I think this is still planned, so I'm going to leave this ticket open for that.

benmay:
I think that would be better done via a constant so it could be deployed to production environments easier? Rather than having to run wp-cli commands on prod server(s).

In our case this week, we moved 30GB of uploads via s3cmd, then did a SQL update on the post_content to map all the urls, then simply deployed the drop in to take over from there -- so enabling wouldn't really have been beneficial - but I'm sure there are other edge cases :)

PR #28, opened 2015-06-22
Enable the plugin via CLI

ericmann:
Establishes a new constant and two CLI commands for enabling/disabling the plugin's URL rewriting mechanism.

If the constant `S3_UPLOADS_AUTOENABLE` is undefined or defined and set to true, the plugin behaves as usual. If it's defined and set to `false`, then the plugin will be active, but image URLs won't automatically point to S3 until explicitly enabled.

To enable the URL rewriting, you'd fire the `wp s3-uploads enable` command that's already documented in the readme. This command will set an autoloaded option in WordPress which is checked the same time as the constant above.

So, briefly:

If the constant is undefined
... or if the constant is defined and set to true
... or if the constant is defined and set to false and the WordPress option exists and is set to 'enabled'
then the plugin will behave entirely as it does today.

The constant gives existing sites the ability to install and activate the plugin, upload media to S3, and explicitly turn on the front-end features of the system without any impact on visitors' experience with the site.

Fixes humanmade/S3-Uploads#15

### DIFF
--- a/inc/class-s3-uploads-wp-cli-command.php
+++ b/inc/class-s3-uploads-wp-cli-command.php
@@ -365,9 +365,7 @@ class S3_Uploads_WP_CLI_Command extends WP_CLI_Command {
 	 * Ensable the auto-rewriting of media links to S3
 	 */
 	public function enable( $args, $assoc_args ) {
-		if ( false === add_option( 's3_uploads_enabled', 'enabled', '', 'yes' ) ) {
-			update_option( 's3_uploads_enabled', 'enabled' );
-		}
+		update_option( 's3_uploads_enabled', 'enabled' );
 
 		WP_CLI::success( 'Media URL rewriting enabled.' );
 	}
--- a/inc/class-s3-uploads-wp-cli-command.php
+++ b/inc/class-s3-uploads-wp-cli-command.php
@@ -361,6 +361,26 @@ class S3_Uploads_WP_CLI_Command extends WP_CLI_Command {
 		WP_CLI::success( sprintf( 'Successfully deleted %s', $prefix ) );
 	}
 
+	/**
+	 * Ensable the auto-rewriting of media links to S3
+	 */
+	public function enable( $args, $assoc_args ) {
+		if ( false === add_option( 's3_uploads_enabled', 'enabled', '', 'yes' ) ) {
+			update_option( 's3_uploads_enabled', 'enabled' );
+		}
+
+		WP_CLI::success( 'Media URL rewriting enabled.' );
+	}
+
+	/**
+	 * Disable the auto-rewriting of media links to S3
+	 */
+	public function disable( $args, $assoc_args ) {
+		delete_option( 's3_uploads_enabled' );
+
+		WP_CLI::success( 'Media URL rewriting disabled.' );
+	}
+
 	private function recurse_copy($src,$dst) {
 		$dir = opendir($src);
 		@mkdir($dst);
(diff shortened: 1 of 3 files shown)


inc/class-s3-uploads-wp-cli-command.php
```diff
@@ -361,6 +361,26 @@ public function rm( $args, $args_assoc ) {
 		WP_CLI::success( sprintf( 'Successfully deleted %s', $prefix ) );
 	}
 
+	/**
+	 * Ensable the auto-rewriting of media links to S3
+	 */
+	public function enable( $args, $assoc_args ) {
+		if ( false === add_option( 's3_uploads_enabled', 'enabled', '', 'yes' ) ) {
```
joehoyle: Why is this attempting to add the option? Can it not just do `update_option( 's3_uploads_enabled', 'enabled' );`

s3-uploads.php
```diff
@@ -21,6 +21,13 @@ function s3_uploads_init() {
 		return;
 	}
 
+	// Make sure the plugin is enabled when autoenable is on, or in a CLI system (so the migrate command works)
+	if ( ( ! defined( 'WP_CLI') || ! WP_CLI ) &&                                      // If CLI is enabled, skip our other checks
```
joehoyle: In the case where someone has `enabled` off, and they do something in WP CLI, this is going to have the unintended affect of enabling s3-uploads when it shouldn't be. 

A better approach I think would be to do 

```PHP
$instance = S3_Uploads::get_instance();
$instance->register_stream_wrapper();
```

In the wp-cli s3-uploads command in the case of it being disabled.

tests/bootstrap.php
```diff
@@ -21,6 +21,9 @@
 require $test_root . '/includes/functions.php';
 
 function _manually_load_plugin() {
+	// Make sure the plugin is "enabled"
+	define( 'S3_UPLOADS_AUTOENABLE', true );
```
joehoyle: From what I understand, this is not a required constant, in which case I would rather run the unit tests without it defined as per the majority use case, and lets us write unit tests specifically with it defined.

joehoyle:
@ericmann thanks for the PR! Sorry for the delay in getting back - couple of notes here.

inc/class-s3-uploads-wp-cli-command.php
ericmann: Functionally, yes. The point of attempting to use `add_option()` first is to explicitly set the autoload argument to `yes`. That's not an argument for `update_option()`. While `update_option()` will default to autoloading if an option does not exist _now_, there's no guarantee it will do so in the future and we want to be very clear that the autoloading of `s3_uploads_enabled` here is intentional and not an artifact of updating a non-existent option.

s3-uploads.php
ericmann: We need to make sure the plugin is "enabled" when CLI is being used, otherwise the migration routines will not fire. I'm not sure how your better approach would work in this case ...

tests/bootstrap.php
ericmann: That's perfectly fine, and I can remove this. I was setting the constant to `true` in attempts to get Travis to behave properly when running CI tests (see #31). The tests on Travis are apparently running without _any_ of the S3 constants defined when I push the patch, which causes the tests to fail entirely as the inclusion of various classes never fires.

inc/class-s3-uploads-wp-cli-command.php
rmccue: I strongly doubt that the default will ever change, as this would break backwards compatibility.

s3-uploads.php
rmccue: Can we late-enable the plugin when the command is run for the CLI? This definitely needs to be simplified down from a giant conditional to multiple returns if the checks don't change too, multiline conditionals are terribly hard to read.

ericmann:
OK, I've made some changes based on your feedback. I've also set some environment variables locally and can confirm the tests look OK on this end, even if Travis has trouble running them remotely.

PR #29, opened 2015-06-27
Support for using the ec2 instance profile

joehoyle:
Support for a S3_UPLOADS_USE_INSTANCE_PROFILE constant to use the instance
profile form the ec2 machine


inc/class-s3-uploads.php
```diff
@@ -17,7 +17,13 @@ class S3_Uploads {
 	public static function get_instance() {
 
 		if ( ! self::$instance ) {
-			self::$instance = new S3_Uploads( S3_UPLOADS_BUCKET, S3_UPLOADS_KEY, S3_UPLOADS_SECRET, defined( 'S3_UPLOADS_BUCKET_URL' ) ? S3_UPLOADS_BUCKET_URL : null, defined( 'S3_UPLOADS_REGION' ) ? S3_UPLOADS_REGION : null );
+
+			$key    = defined( 'S3_UPLOADS_KEY' ) ? S3_UPLOADS_KEY : null;
```
spacedmonkey: Can we add some filters here please :smile:

spacedmonkey:
Does this break the use of keys and force you to use instance profiles? If so, don't you have to be in AWS to use them? What if you are hosting your images in s3 but are using say, digital ocean?

joehoyle:
@spacedmonkey no this is if you _want_ to use the instance profile, the keys thing is unchanged.

inc/class-s3-uploads.php
joehoyle: Added in be6fa1e

Issue #12, opened 2015-06-01
Fix Image Magick


joehoyle:
Because image magick uses a library binding, it uses the local filesystem under the hood, which appears to be an issue with using S3-Uploads. We should somehow work around this, by subclassing `WP_Image_Magick_Editor` or otherwise.

joehoyle:
Fixed in #14!

PR #33, opened 2015-07-04
Enable the plugin via CLI

joehoyle:
See #12 #28


joehoyle:
@ericmann I created an `enabled` function to more easily to the detection if it's enabled/disabled, and split the setup stuff so I could add some tests.

ericmann:
Looks good to me.

PR #36, opened 2015-07-27
Add proxy information to AWS SDK config

eduardoboucas:
The plugin silently fails when used behind a proxy, as it doesn't pass the proxy settings from `wp-config.php` to the AWS SDK.

To get around this, I'm currently using a small plugin that intercepts the `s3_uploads_s3_client_params` filter, checks if `WP_PROXY_HOST` and `WP_PROXY_PORT` are defined and, if so, add that information to the SDK configuration [as per the documentation](http://docs.aws.amazon.com/aws-sdk-php/v2/guide/configuration.html#using-a-proxy).

It would be great to not require an additional plugin for this, so I've included that logic in this PR in case you want to include it in the core.

Thanks!

### DIFF
--- a/inc/class-s3-uploads.php
+++ b/inc/class-s3-uploads.php
@@ -147,7 +147,7 @@ class S3_Uploads {
 			$proxy_address = WP_PROXY_HOST . ':' . WP_PROXY_PORT;
 
 			if ( defined('WP_PROXY_USERNAME') && defined('WP_PROXY_PASSWORD') ) {
-				$proxy_auth = WP_PROXY_USERNAME . ':' . WP_PROXY_USERNAME . '@';
+				$proxy_auth = WP_PROXY_USERNAME . ':' . WP_PROXY_PASSWORD . '@';
 			}
 
 			$params['request.options']['proxy'] = $proxy_auth . $proxy_address;
--- a/inc/class-s3-uploads.php
+++ b/inc/class-s3-uploads.php
@@ -142,6 +142,10 @@ class S3_Uploads {
 			$params['region'] = $this->region;
 		}
 
+		if ( defined('WP_PROXY_HOST') && defined('WP_PROXY_PORT') ) {
+			$params['request.options']['proxy'] = WP_PROXY_HOST . ':' . WP_PROXY_PORT;
+		}
+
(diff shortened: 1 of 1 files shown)


spacedmonkey:
This is what we [currently doing](https://github.com/TimeIncUK/wp-amazon-web-services/blob/master/classes/amazon-web-services.php#L170-L173) in our S3 plugin. Really all communication should be going through wp_remote_get / wp_remote_post, as these are filterable etc. But in this case, I think it going to be more pain that it is worth to hack all the calls to use these functions. The defines @eduardoboucas is using are WP standard, as seen [here](https://github.com/WordPress/WordPress/blob/master/wp-includes/class-http.php#L1695). 

Good work @eduardoboucas :+1:

willmot:
Assigned to @joehoyle for a review/merge, thanks for the contribution!

rmccue:
Hmm, tests are failing here with "Class 'S3_Uploads_Image_Editor_Imagick' not found", which seems to be unrelated to the PR.

joehoyle:
This is because I have to pull the branches to this remote so tracks gets the secrets

inc/class-s3-uploads.php
```diff
@@ -142,6 +142,17 @@ public function s3() {
 			$params['region'] = $this->region;
 		}
 
+		if ( defined('WP_PROXY_HOST') && defined('WP_PROXY_PORT') ) {
+			$proxy_auth = '';
+			$proxy_address = WP_PROXY_HOST . ':' . WP_PROXY_PORT;
+
+			if ( defined('WP_PROXY_USERNAME') && defined('WP_PROXY_PASSWORD') ) {
+				$proxy_auth = WP_PROXY_USERNAME . ':' . WP_PROXY_USERNAME . '@';
```
danielhomer: Shouldn't this be
    
    $proxy_auth = WP_PROXY_USERNAME . ':' . WP_PROXY_PASSWORD . '@';

eduardoboucas: You're absolutely right. Fixed! Thanks @danielhomer.

joehoyle:
Moved this to #43 with a cleanup commin

joehoyle:
@eduardoboucas thanks very much! Merged in #43

Issue #44, opened 2015-08-12
CLI / Migrating attachments / stream_wrapper_register(): Protocol s3:// is already defined


rob:
I'm running `wp s3-uploads migrate-attachments` and get the above-mentioned error after every 10 uploads:

```
$ wp s3-uploads migrate-attachments
Success: Moved file 2015/08/Slider-FULL-EPISODE.jpg to S3
Success: Moved file 2015/08/Slider-FULL-EPISODE-150x150.jpg to S3
Success: Moved file 2015/08/Slider-FULL-EPISODE-300x189.jpg to S3
Success: Moved file 2015/08/Slider-FULL-EPISODE-233x146.jpg to S3
Success: Moved file 2015/08/Slider-FULL-EPISODE-341x214.jpg to S3
Success: Moved file 2015/08/Slider-FULL-EPISODE-600x375.jpg to S3
Success: Moved file 2015/08/Slider-FULL-EPISODE-700x438.jpg to S3
Success: Moved file 2015/08/Slider-FULL-EPISODE-140x88.jpg to S3
Success: Moved file 2015/08/Slider-FULL-EPISODE-700x241.jpg to S3
PHP Warning:  stream_wrapper_register(): Protocol s3:// is already defined. in /home/www/wp-content/plugins/S3-Uploads-master/inc/class-s3-uploads-stream-wrapper.php on line 11
Warning: stream_wrapper_register(): Protocol s3:// is already defined. in /home/www/wp-content/plugins/S3-Uploads-master/inc/class-s3-uploads-stream-wrapper.php on line 11
Success: Moved file 2015/07/Web-Article.jpg to S3
Success: Moved file 2015/07/Web-Article-150x150.jpg to S3
Success: Moved file 2015/07/Web-Article-300x189.jpg to S3
Success: Moved file 2015/07/Web-Article-1024x644.jpg to S3
Success: Moved file 2015/07/Web-Article-233x146.jpg to S3
Success: Moved file 2015/07/Web-Article-341x214.jpg to S3
Success: Moved file 2015/07/Web-Article-600x375.jpg to S3
Success: Moved file 2015/07/Web-Article-700x438.jpg to S3
Success: Moved file 2015/07/Web-Article-140x88.jpg to S3
Success: Moved file 2015/07/Web-Article-1140x241.jpg to S3
[...]

PR #45, opened 2015-08-22
unregister the 's3' stream wrapper if it's already registered

rob:
See #44.

Prevents `Warning: stream_wrapper_register(): Protocol s3:// is already defined.` messages from popping up when run via cli (e.g., `wp s3-uploads migrate-attachments`.)

### DIFF
--- a/inc/class-s3-uploads.php
+++ b/inc/class-s3-uploads.php
@@ -67,6 +67,10 @@ class S3_Uploads {
 	 */
 	public function register_stream_wrapper() {
 		if ( defined( 'S3_UPLOADS_USE_LOCAL' ) && S3_UPLOADS_USE_LOCAL ) {
+			if ( in_array( 's3', stream_get_wrappers() ) ) {
++                		stream_wrapper_unregister( 's3' );
++          		}
+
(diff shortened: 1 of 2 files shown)


joehoyle:
Thanks @rob I'll take a look. Also, you must be an early GitHub user with that handle!

rob:
Thanks @joehoyle. And yeah, but it gets annoying when people use @rob for shorthand in public and I get notifications all the time!

inc/class-s3-uploads.php
```diff
@@ -67,6 +67,10 @@ public function tear_down() {
 	 */
 	public function register_stream_wrapper() {
 		if ( defined( 'S3_UPLOADS_USE_LOCAL' ) && S3_UPLOADS_USE_LOCAL ) {
+			if ( in_array( 's3', stream_get_wrappers() ) ) {
++                		stream_wrapper_unregister( 's3' );
++          		}
```
bdurette: I think these pluses are an artifact of a copy/paste from a patch and are the likely cause of tests failing.

joehoyle:
This was fixed in #48 so should be able to close this out now.

PR #52, opened 2015-09-17
Added ability to override s3 url.

eugene-manuilov:


joehoyle:
I think the filter name should be `s3_uploads_bucket_url` would be better, `s3_url` is too generic.

eugene-manuilov:
Updated filter name.

fklein-lu:
Marking that the test failure is unrelated to the PR.

PR #59, opened 2015-11-03
Update README.md

ergunozyurt:
included S3_Uploads_Region definition example for Frankfurt and Beijing regions which use Signature v4 for authentication.


joehoyle:
Fixed in f39078a thanks!

Issue #62, opened 2015-11-16
Unit Tests are failing


fklein-lu:
When running unit tests, there are two errors:
```
FFPHP Fatal error:  Class 'S3_Uploads_Image_Editor_Imagick' not found in /srv/www/s3-uploads/htdocs/wp-content/plugins/s3-uploads/tests/test-s3-uploads-image-editor-imagick.php on line 46

Fatal error: Class 'S3_Uploads_Image_Editor_Imagick' not found in /srv/www/s3-uploads/htdocs/wp-content/plugins/s3-uploads/tests/test-s3-uploads-image-editor-imagick.php on line 46
```

fklein-lu:
This was fixed in #63, closing.

PR #63, opened 2015-11-17
Add autoloader for plugin classes.

fklein-lu:
This avoids PHP fatal errors in unit tests, as those not always include
the necessary classes.

See #62.

### DIFF
--- a/s3-uploads.php
+++ b/s3-uploads.php
@@ -51,3 +51,26 @@ function s3_uploads_enabled() {
 
 	return true;
 }
+
+/**
+ * Autoload callback.
+ *
+ * @param $class_name Name of the class to load.
+ */
+function s3_uploads_autoload( $class_name ) {
+	/*
+	 * Load plugin classes:
+	 * - Class name: S3_Uploads_Image_Editor_Imagick.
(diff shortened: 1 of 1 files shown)


mattheu:
I like the autoloader - but does this mean we can also drop the use of `require_once` at the top of the file to include 

`dirname( __FILE__ ) . '/inc/class-s3-uploads.php';`
and
`dirname( __FILE__ ) . '/inc/class-s3-uploads-wp-cli-command.php';`

s3-uploads.php
```diff
@@ -51,3 +51,26 @@ function s3_uploads_enabled() {
...
+	 * - Class name: S3_Uploads_Image_Editor_Imagick.
+	 * - File name: class-s3-uploads-image-editor-imagick.php.
+	 */
+	$class_file = 'class-' . strtolower( str_replace( '_', '-', $class_name ) ) . '.php';
+	$class_path = dirname( __FILE__ ) . '/inc/' . $class_file;
+
+	if ( file_exists( $class_path ) ) {
+		require $class_path;
+
+		return;
```
mattheu: Do we need this?

mattheu:
Chatted about further use of the autoloader - can tackle that separately if we want.

PR #64, opened 2015-11-17
Remove require statements that are no longer needed.

fklein-lu:
An autoloader for the plugin classes was introduced in d57dec0.

This change means that these class files do not have to be required
manually anymore.


mattheu:
looks good :+1:

PR #75, opened 2015-11-19
Check for PHP version requirements

### LINKED ISSUE #69: Check for required PHP version
The plugin requires at least PHP 5.3.3 (edit: adapted PHP version to that of SDK V2), following the requirements of the AWS SDK. This is higher than the WordPress requirements.

We should add a check to the plugin that stops the execution when the minimum is not met, to avoid a fatal PHP error, and add a notice to the documentation.

fklein-lu:
Add a notice when the requirements are not met, and return to avoid
executing the rest of the plugin.

Fixes #69.

### DIFF
--- a/s3-uploads.php
+++ b/s3-uploads.php
@@ -15,6 +15,10 @@ if ( defined( 'WP_CLI' ) && WP_CLI ) {
 add_action( 'plugins_loaded', 's3_uploads_init' );
 
 function s3_uploads_init() {
+	if ( ! s3_uploads_check_requirements() ) {
+		return;
+	}
+
 	if ( ! defined( 'S3_UPLOADS_BUCKET' ) ) {
 		return;
 	}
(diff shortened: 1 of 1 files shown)


mattheu:
This looks pretty neat. One thought - could we just call `wp_die` if PHP version is too low? I think the WP plugin loader is smart enough to handle that, not load the plugin and display a notice.

fklein-lu:
`wp_die()` by default calls `_default_wp_die_handler()`, which calls `die()`, which kills the entire script execution.

mattheu:
OK cool - lets go with this.

s3-uploads.php
```diff
@@ -32,6 +36,34 @@ function s3_uploads_init() {
...
+	return true;
+}
+
+/**
+ * Print an admin notice when the PHP version is not high enough.
+ *
+ * This has to be a named function for compatibility with PHP 5.2.
+ */
+function s3_uploads_outdated_php_version_notice() {
+	printf( '<div class="error"><p>The S3 Uploads plugin requires PHP version 5.3.3 or higher. Your server is running PHP version %s.</p></div>',
```
mattheu: This should probably be translatable.

mattheu:
One comment on making this translatable - but otherwise good to merge.

s3-uploads.php
```diff
@@ -32,6 +36,34 @@ function s3_uploads_init() {
 }
 
 /**
+ * Check whether the environment meets the plugin's requirements, like the minimum PHP version.
+ *
+ * @return bool True if the requirements are met, else false.
+ */
+function s3_uploads_check_requirements() {
+	if ( version_compare( '5.3.3', PHP_VERSION, '>=' ) ) {
```
joehoyle: From what I can gather, this is going to be true if `PHP_VERSION` is `5.3.3` which we don't want. It should be using  `>`.

fklein-lu: Saw that today, and fixed it in d2e6485.

Issue #39, opened 2015-07-28
`verify` CLI command successful despite S3_ constants not defined


rob:
I'm testing S3-Uploads (master branch) on a server. I did not define S3_UPLOADS_BUCKET, S3_UPLOADS_KEY, or S3_UPLOADS_SECRET in `wp-config.php` yet and tried to run `wp s3-uploads verify`:

```
cptv@nj:~/www$ wp s3-uploads verify
Success: Looks like your configuration is correct.
Attempting to upload file /home/[...]/www/wp-content/uploads/484304978.jpg
File uploaded to S3 successfully
Attempting to delete file /home/[...]/www/wp-content/uploads/484304978.jpg
File deleted from S3 successfully

cptv@nj:~/www$ wp s3-uploads ls
Error: Error retrieving credentials from the instance profile metadata server. When you are not running inside of Amazon EC2, you must provide your AWS access key ID and secret access key in the "key" and "secret" options when creating a client or provide an instantiated Aws\Common\Credentials\CredentialsInterface object. ([curl] 28: Connection timed out after 5006 milliseconds [url] http://169.254.169.254/latest/meta-data/iam/security-credentials/)
```

Shouldn't `wp s3-uploads verify` be failing as there's no credentials listed anywhere?

joehoyle:
Oops, looks like verify is running regardless, so just basically testing files locally is working, which doesn't make much sense. Thanks for the report!

rob:
No worries.

(I'm the Rob you talked to the other day with the mail fail issue. Glad I could find a bug already :))

PR #76, opened 2015-11-20
Adapt `verify` command for Instance Profiles.

fklein-lu:
When using AWS Instance Profiles, no credentials (Key ID, Secret Key)
need to be set.

See #39.


joehoyle:
Looks good

PR #81, opened 2015-12-20
Fix #69 Disable plugin if requirement not match

### LINKED ISSUE #69: Check for required PHP version

PatelUtkarsh:
Should also disabled plugin if it is not matching requirement.

### DIFF
--- a/s3-uploads.php
+++ b/s3-uploads.php
@@ -36,6 +36,13 @@ function s3_uploads_init() {
 }
 
 /**
+ * Disable self
+ */
+function _s3_deactivate_self() {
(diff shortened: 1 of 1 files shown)


joehoyle:
Personally I don't think we should disable the plugin in this case, I'm fine with it just showing the error message and returning.

PR #85, opened 2016-01-31
Missing installation step added to README.md file.

matthewfedak:
After adding the plugin files to your WordPress install you must activate the plugin before you can use the s3-uploads command line tools. 

I have added an extra step to the README.md file wit the WP-CLI activation command.


joehoyle:
Thanks!

PR #88, opened 2016-03-31
Automation output improvement

dvaness:
We are attempting to use this plugin for rolling out S3 Uploads across ~700 sites. Therefore, we need to be able to automate the installation process. This change will add a new WP CLI argument for the create-iam-user action which will result in a JSON encoded string being printed instead of the user friendly output that currently exists.

### DIFF
--- a/inc/class-s3-uploads-wp-cli-command.php
+++ b/inc/class-s3-uploads-wp-cli-command.php
@@ -139,7 +139,7 @@ class S3_Uploads_WP_CLI_Command extends WP_CLI_Command {
 	 * Create an AWS IAM user for S3 Uploads to user
 	 *
 	 * @subcommand create-iam-user
-	 * @synopsis --admin-key=<key> --admin-secret=<secret> [--username=<username>] [--automatic]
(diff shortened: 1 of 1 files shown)


inc/class-s3-uploads-wp-cli-command.php
```diff
@@ -139,7 +139,7 @@ public function migrate_attachment_to_s3( $args, $args_assoc ) {
 	 * Create an AWS IAM user for S3 Uploads to user
 	 *
 	 * @subcommand create-iam-user
-	 * @synopsis --admin-key=<key> --admin-secret=<secret> [--username=<username>]
+	 * @synopsis --admin-key=<key> --admin-secret=<secret> [--username=<username>] [--automatic]
```
danielbachhuber: What about `--format=json` instead of `--automatic`, for better parity with other WP-CLI commands?

dvaness: That would work. Would you also want support for other output styles?

joehoyle:
I cleanup up the formatting in #89, thanks a lot for the patch!

PR #90, opened 2016-03-31
Run tests under php7

joehoyle:


larssn:
We've experienced no php-specific problems on php7.06. Been running it for over a month.

joehoyle:
Code elsewhere!

PR #97, opened 2016-05-06
Allow bucket URL to be filtered even when defined

sanchothefat:
Understand this might be intentionally unfiltered so the constant is definitely constant. This just allows me to run `set_url_scheme()` on for an edge case on our hosting stack.

Worth adding `set_url_scheme()` in here anyway instead of the filter?


sanchothefat:
@joehoyle closing this - wasn't thinking it through :)

PR #101, opened 2016-05-24
Remove lib/aws-sdk, load in via composer instead

danielhomer:
Thanks to the work carried out by porting to v3 of the AWS SDK and the addition of the `S3_Uploads_Stream_Wrapper` class, we can now ditch manually including the SDK in the `lib` dir. A couple of advantages of doing this:
- Our WordPress environment is built with composer and we have other plugins which require the AWS SDK. By removing the SDK and loading it in by composer instead, we can handle the dependency ourselves. 
- No more huge diffs 😄 

For users not using composer to build their WP environments, all that's needed is a `composer install` in the plugin dir as we now check for `vendor/autoload.php` in `s3-uploads.php`.


danielhomer:
This requires a lot more work than I previously thought, tests pass but uploads aren't working in UAT, will loop back on this again soon.

joehoyle:
Hmm interesting, not sure why this would have broken anything - did you inadvertently change the version of the SDK perhaps?

Thought I detest composer, I think I'm ok with this, as I know having lots of versions of the AWS SDK bundled in different plugins can be a total pain. Also, we ship releases of this plugin to GitHub, we'll just need to change that process to bundle the AWS SDK in the ZIP I'd imagine.

Tentatively, I think this could be adopted.

danielhomer:
Managed to get a bit more time on this today and got a working example working in UAT. Turns out the issue was with the constructor call at https://github.com/humanmade/S3-Uploads/blob/master/inc/class-s3-uploads.php#L22-L23. These constants aren't defined in our environment (using IAM) so these were getting passed through to the credentials as strings, not null, causing 403 errors further down the line as the SDK was trying to authenticate with them. I've put conditionals on all of these params now.

I've also replaced the `S3_Uploads_Stream_Wrapper::register()` call in favour of `Aws\S3\StreamWrapper::register()` - I'd be interested to know if there was a specific reason the AWS wrapper wasn't being used, though.

That's about it so far, a few bits still need tidying up, but let me know what you think.

rmccue:
> I'd be interested to know if there was a specific reason the AWS wrapper wasn't being used, though.

I believe this is so we can implement local mode for testing.

danielhomer:
@rmccue Is this a requirement? In my opinion, the plugin should be lightweight as possible and the stream logic should be left to the AWS SDK.

joehoyle:
We use a custom stream wrapper to work around a couple of core issues too (such as a `is_dir()` check on every page load via `wp_upload_dir`. Also, we had to patch it for some MIME type detection bugs in the AWS SDK, though those may have been fixed in later versions.

danielhomer:
The `$create_dir` parameter was added to `wp_upload_dir` in 4.5, so that appears to have fixed the `is_dir()` problem, although that does severely impact the backwards compatibility of the plugin. 

As for the MIME type issue, that seems to have been fixed in the AWS SDK (https://github.com/aws/aws-sdk-php/blob/master/src/S3/StreamWrapper.php#L169-L175).

sc0ttkclark:
+1 on this, we really need this. Because if you are using any of the libraries in any other project you include that utilizes Composer, the autoloaders conflict :(

sc0ttkclark:
Example, using the `elasticsearch/elasticsearch` project in composer, it includes GuzzleHTTP libraries and has the autoloader etc. It conflicts pretty harshly:

`Fatal error: Cannot redeclare GuzzleHttp\json_encode() (previously declared in /path/to/plugin/vendor/guzzlehttp/guzzle/src/functions.php:322) in /path/to/s3-uploads/lib/aws-sdk/GuzzleHttp/functions.php on line 321`

rmccue:
@sc0ttkclark Since #172, if the `Aws\S3\S3Client` is already defined, it won't attempt to load the SDK again. That should already solve the Composer conflict problem as far as I can see?

sc0ttkclark:
That will work, I'll try loading it via composer ahead of time, that way it won't conflict with what the SDK itself uses.

sc0ttkclark:
That worked, thanks!

joehoyle:
This has sinse been done, thanks for paving the way @danielhomer

PR #116, opened 2016-08-08
Handle wp_read_image_metadata

### LINKED ISSUE #115: exif_read_data() doesn't support file streams
Images at the point of the `media_handle_upload` filter will be on S3 and as a result subsequent image handling in particular `exif_read_data()` called during image metadata generation returns false. This isn't ideal if we need to do anything with orientation data for example, it always defaults to 0.

roborourke:
Fixes #115

Could do with someone else running the tests first just to confirm for me. With and without the filter in `class-s3-uploads.php`.


inc/class-s3-uploads.php
```diff
@@ -208,12 +209,108 @@ public function filter_editors( $editors ) {
...
+	/**
+	 * Filters wp_read_image_metadata. exif_read_data() doesn't work on
+	 * file streams so we need to make a temporary local copy to extract
+	 * exif data from.
+	 *
+	 * @param array  $meta Array of metadata
+	 * @param string $file File name on S3
+	 * @return array
+	 */
+	public function wp_filter_read_image_metadata( $meta, $file ) {
```
joehoyle: Rather than copy the entire function, would we move the filter to our own function and call it with the copied file:

```PHP
remove_filter( ...the old filter );
add_filter( filter, array( $this, 'replaced_function' ) );

function replaced_function( $file ) {
    $new_file = copy_file_from_s3( $file );
    $data = old_function( $new_file );
    unlink( $file );
    return $data;
}
```

roborourke: Good shout, that makes sense

-- 
*Robert O'Rourke*
Senior WordPress Engineer
humanmade

*phone: *+447816329424
*email:* rob@hmn.md
*skype:* rob.o.rourke
*twitr:* @sanchothefat <https://twitter.com/sanchothefat>

roborourke: @joehoyle Not sure if this is exactly what you meant. The original function is called directly and not through `apply_filters()` so had to remove ours before the function call and re-add it after. I may have misunderstood though.

joehoyle: Ahh crap, I thought it was called via a filter (on wp_handle_upload) or something.

inc/class-s3-uploads.php
```diff
@@ -208,12 +209,42 @@ public function filter_editors( $editors ) {
...
+	 * file streams so we need to make a temporary local copy to extract
+	 * exif data from.
+	 *
+	 * @param array  $meta
+	 * @param string $file
+	 * @return array|bool
+	 */
+	public function wp_filter_read_image_metadata( $meta, $file ) {
+		remove_filter( 'wp_read_image_metadata', array( $this, 'wp_filter_read_image_metadata' ), 10, 2 );
+		$temp_file = $this->copy_image_from_s3( $file );
```
joehoyle: You seem to be missing the `callback` param, also see other warnings in https://travis-ci.org/humanmade/S3-Uploads/jobs/150781407

roborourke:
ugh, shouldn't code after midnight! I was thinking of passing a callback to the temp copy function so the copying and unlinking could be handled in the one place but it's not as flexible that way.

PR #120, opened 2016-08-31
Set minimum required PHP version to 5.5.0

NathanKleekamp:
This pull request sets the minimum PHP version to 5.5.0, which is the minimum PHP version required by the AWS-SDK v3. It threw a fatal error when I tried to activate on a machine running 5.4.*


joehoyle:
Thanks!

PR #128, opened 2016-09-23
Lazy load the S3Client() for performance

joehoyle:
Currently, the S3 Client is initialized on every requests, when the
stream wrapper is registered. This adds about 20ms - 30ms to every
request, whether the custom stream will be used or not. Instead, we can
lazy load the S3 Client to only initalize it once it's used.

### DIFF
--- a/inc/class-s3-uploads-stream-wrapper.php
+++ b/inc/class-s3-uploads-stream-wrapper.php
@@ -95,15 +95,17 @@ class S3_Uploads_Stream_Wrapper
 	/** @var string The opened protocol (e.g., "s3") */
 	private $protocol = 's3';
 
+	private $client;
+
 	/**
 	 * Register the 's3://' stream wrapper
 	 *
-	 * @param S3ClientInterface $client   Client to use with the stream wrapper
+	 * @param array             $client_config   Client to use with the stream wrapper
 	 * @param string            $protocol Protocol to register as.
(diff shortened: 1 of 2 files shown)


inc/class-s3-uploads-stream-wrapper.php
```diff
@@ -95,15 +95,17 @@ class S3_Uploads_Stream_Wrapper
 	/** @var string The opened protocol (e.g., "s3") */
 	private $protocol = 's3';
 
+	private $client;
```
rmccue: Needs phpDoc

inc/class-s3-uploads-stream-wrapper.php
```diff
@@ -725,11 +727,17 @@ private function getOption($name)
 	 */
 	private function getClient()
 	{
-		if (!$client = $this->getOption('client')) {
+		if( $this->client ) {
```
rmccue: Spacing after `if`

inc/class-s3-uploads.php
```diff
@@ -154,11 +154,16 @@ public function get_original_upload_dir() {
...
 		if ( ! empty( $this->s3 ) ) {
 			return $this->s3;
 		}
 
+		$this->s3 = Aws\S3\S3Client::factory( $this->get_s3_config() );
+
+		return $this->s3;
+	}
+
+	public function get_s3_config() {
```
rmccue: Needs phpDoc

Issue #95, opened 2016-04-29
Where is "Aws\Common\Aws" ?


twmbx:
Have I done something wrong?
Just installed S3-Uploads & can't get verify or create-iam-user to work because of

```
Fatal error: Uncaught Error: Class 'Aws\Common\Aws' not found
```

When I checked the [aws sdk autoloader](https://github.com/humanmade/S3-Uploads/blob/master/lib/aws-sdk/aws-autoloader.php), I can't seem to find a mapping for "Aws\Common\Aws" ... Just me?

spacedmonkey:
Interesting I am getitng this 

```
PHP Fatal error:  Class 'Aws\\Common\\Exception\\ServiceResponseException'
```

I wonder if it is related...

twoSeats:
@spacedmonkey i would expect so, there are no "Aws\Common" mappings at all from what I could see.

joehoyle:
@twoSeats can you make sure you are using the latest stable release from https://github.com/humanmade/S3-Uploads/releases thanks!

twoSeats:
@joehoyle oh is that what it is? I'll give that a shot.

seostudio:
Same problem. Newest version installed.

simonhampel:
Just setting this up for the first time - pulled the latest version, tried to run create-iam-user and had this problem with Aws\Common\Aws missing:

```
$ wp s3-uploads create-iam-user --admin-key=XXX --admin-secret=YYY
PHP Fatal error:  Uncaught Error: Class 'Aws\Common\Aws' not found in /srv/www/example.com/wp-content/plugins/s3-uploads/inc/class-s3-uploads-wp-cli-command.php:155
Stack trace:
#0 [internal function]: S3_Uploads_WP_CLI_Command->create_iam_user(Array, Array)
#1 phar:///usr/local/bin/wp/php/WP_CLI/Dispatcher/CommandFactory.php(67): call_user_func(Array, Array, Array)
#2 [internal function]: WP_CLI\Dispatcher\CommandFactory::WP_CLI\Dispatcher\{closure}(Array, Array)
#3 phar:///usr/local/bin/wp/php/WP_CLI/Dispatcher/Subcommand.php(351): call_user_func(Object(Closure), Array, Array)
#4 phar:///usr/local/bin/wp/php/WP_CLI/Runner.php(327): WP_CLI\Dispatcher\Subcommand->invoke(Array, Array, Array)
[...]

PR #133, opened 2016-10-17
Fixes compatibility issues with aws-php-sdk v3.

puckbag:
- Update README about `create-iam-user`. More details about AWS environment credentials and profiles would be nice.
- `Aws\Common\Aws` no longer exists.
- Replace deprecated `::factory` calls with proper instantiations.

### DIFF
--- a/inc/class-s3-uploads.php
+++ b/inc/class-s3-uploads.php
@@ -7,6 +7,7 @@ class S3_Uploads {
 	private        $bucket_url;
 	private        $key;
 	private        $secret;
+    private        $s3;
 
 	public $original_upload_dir;
 
@@ -183,7 +184,7 @@ class S3_Uploads {
 		}
 
 		$params   = apply_filters( 's3_uploads_s3_client_params', $params );
-		$this->s3 = Aws\S3\S3Client::factory( $params );
+		$this->s3 = new Aws\S3\S3Client( $params );
 
 		return $this->s3;
 	}
--- a/inc/class-s3-uploads.php
+++ b/inc/class-s3-uploads.php
@@ -7,7 +7,7 @@ class S3_Uploads {
 	private        $bucket_url;
 	private        $key;
 	private        $secret;
-    private        $s3;
+	private        $s3;
 
 	public $original_upload_dir;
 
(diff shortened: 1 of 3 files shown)


inc/class-s3-uploads.php
```diff
@@ -7,6 +7,7 @@ class S3_Uploads {
 	private        $bucket_url;
 	private        $key;
 	private        $secret;
+    private        $s3;
```
rmccue: Indentation here is a bit off; make sure you're using tabs instead of spaces.

BandonRandon:
Just commenting to say that on my server running Debian Jesse 8 and PHP 7.1 I had to apply this PR/Patch to fix the following fatal:
```PHP Fatal error:  Uncaught TypeError: Argument 1 passed to S3_Uploads_Stream_Wrapper::register() must be an instance of Aws\S3\S3ClientInterface, instance of Aws\S3\S3Client given, called in plugins/s3-uploads/inc/class-s3-uploads.php on line 75 and defined in plugins/s3-uploads/inc/class-s3-uploads-stream-wrapper.php:105
```

This patch cleared things right up for me.
I also wanted to mention I had the same problem noted in  https://github.com/humanmade/S3-Uploads/issues/95#issuecomment-254318557 I was able to fix the conflict by copying the AWS SDK from S3-Uploads to BackupWordPress S3. I reported it https://github.com/humanmade/backupwordpress-pro-s3/issues/34 but since they are both HM products wanted to note here as well.

epoplive:
I had to combine the old and new format command line's to get the command to run, but then I get an error <Message>Credential should be scoped to a valid region, not 'us-west-2'. </Message>

hxpro:
Hi @joehoyle 
Please, can you also check `S3_Uploads_ChangedFilesIterator` ? It extends `Aws\S3\Sync\ChangedFilesIterator`. The `Aws\S3\Sync` namespace have been replaced by the `Aws\S3\Transfer` class in aws-php-sdk v3.

joehoyle:
This is now fixed, as we dswitched ot loading the cdk via composer and fix the compat issues mentioned in this PR. Thanks all!

Issue #134, opened 2016-10-18
chmod(): S3_Uploads_Stream_Wrapper::stream_metadata is not implemented


tillkruss:
Uploads are working, but I'm seeing the following error when uploading new files.

```
PHP Warning: POST /wp/wp-admin/async-upload.php
chmod(): S3_Uploads_Stream_Wrapper::stream_metadata is not implemented
```

joehoyle:
We now have this, thanks!

PR #135, opened 2016-10-19
Add missing `stream_metadata()` method

tillkruss:
The PR prevents at PHP Warning triggered by `_wp_handle_upload()` calling `chmod()` on the custom S3 stream wrapper.

See https://github.com/humanmade/S3-Uploads/issues/134.

### DIFF
--- a/inc/class-s3-uploads-stream-wrapper.php
+++ b/inc/class-s3-uploads-stream-wrapper.php
@@ -237,6 +237,11 @@ class S3_Uploads_Stream_Wrapper
 			});
 	}
 
+	public function stream_metadata($path, $option, $value)
+	{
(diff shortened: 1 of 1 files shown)


rmccue:
Per [the PHP docs](http://php.net/manual/en/streamwrapper.stream-metadata.php):

> Returns TRUE on success or FALSE on failure. If option is not implemented, FALSE should be returned. 

Seems like we should be returning false here then?

tillkruss:
Updated 🙂

tillkruss:
Ping.

rmccue:
Apologies, I'd missed this! Thanks for the ping, and the PR.

PR #136, opened 2016-10-27
Subfolder support

tysonlt:
This small change adds subfolder support.

This is useful for us, as we want to store the content for all of our hosted sites in one bucket. 

If the S3_UPLOADS_SUBFOLDER constant is not detected, then there is no change to default behaviour.


joehoyle:
This is actually already supported by passing the path in teh S3_UPLOADS_BUKCET: 

```
define( 'S3_UPLOADS_BUKCET', 'mybucket/subfolder' );
```

tysonlt:
Ah! I didn't know that.

rmccue:
I'm going to close this out as it's already supported, but thanks for the PR anyway! 😄 🌴

tysonlt:
Thanks mate, maybe I'll do another pr to make it more explicit in the docs

rmccue:
@tysonlt Absolutely, would love this to be documented better :)

PR #138, opened 2016-11-01
Add endpoints support, required for s3 compatible services such as Minio

andrewklau:


joehoyle:
@andrewklau thanks for the contribution. Now we have the filter `s3_uploads_s3_client_params` I think it's better for plugins etc to hook in there rather than defining everything in constants. Constants create another piece of global state that I think is better handled via the filter.

andrewklau:
@joehoyle Thanks for the quick response, I did see `s3_uploads_s3_client_params` mentioned in an earlier issue but couldn't find the actual source function.

andrewklau:
Oh I see.. it's a filter.. can't believe I didn't see that!

joehoyle:
All good!

PR #140, opened 2016-11-21
Fixing temporary upload location

davidlawson:
The temporary file currently ends up in the `inc` directory as `s3-uploadsFILENAMErandom.tmp`. This doesn't work very well when the unix permissions are set to read-only for plugins. This moves things to `wp-content/s3-uploads`.

### DIFF
--- a/inc/class-s3-uploads.php
+++ b/inc/class-s3-uploads.php
@@ -243,7 +243,7 @@ class S3_Uploads {
 	 * @return string
 	 */
 	public function copy_image_from_s3( $file ) {
-		$temp_filename = wp_tempnam( $file, 's3-uploads' );
+		$temp_filename = wp_tempnam( $file, WP_CONTENT_DIR . '/s3-uploads/' );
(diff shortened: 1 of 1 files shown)


jdub:
This is an important change.

rmccue:
This explains the error we've been seeing in New Relic...

Any reason we should specify a directory at all? `wp-content/s3-uploads` may also not be writable, and I don't see a reason not to just use `get_temp_dir()` (aka the default).

cc @joehoyle

jdub:
Default temp dir seems reasonable to me.

joehoyle:
We already use `get_temp_dir()`, see https://github.com/humanmade/S3-Uploads/blob/master/inc/class-s3-uploads-image-editor-imagick.php#L20

joehoyle:
Derp sorry scrap that, but what is this `copy_image_from_s3` function all about?

joehoyle:
Oh I see, yeah oops. Let's merge this :) #runningcommentry

PR #146, opened 2017-01-19
Document the format for S3_UPLOADS_BUCKET_URL

### LINKED ISSUE #145: Addition to Documentation on URL Rewrites
Ran into an issue where using a canonical bucket in development/staging kept throwing HTTPS certificate errors and causing all kinds of grief trying to work with media uploads. So decided to rewrite our URLs for development and staging environments to use the path-based format and bypass the certificate issue (we'll add a CNAME designator in production to solve it there). When I set out to rewrite the URLs it wasn't clear what format the S3_UPLOADS_BUCKET_URL constant needed to be in, so a little extra documentation might be helpful. In particular, I wasn't sure if the plugin would prepend a protocol or append a trailing slash. As it turns out, after learning through trial and error, it doesn't do the former but does do the latter. So you need a protocol on the front of your constant but you need to leave off the trailing slash at the end or you'll end up with a double-trailing slash = broken url.

So for example, if you are using a path-based url rewrite, here's how your constant should look:
'S3_UPLOADS_BUCKET_URL' = 'https://s3.amazonaws.com/bucketname'

And for a CNAME URL:
'S3_UPLOADS_BUCKET_URL' = 'https://media.domain.com'

Hope this is helpful to anyone else who might be experiencing similar issues.

rmccue:
Fixes #145.


Japh:
LGTM

PR #160, opened 2017-03-27
add control of S3 object permissions

philipowen:
Allow the S3 object permissions to be controlled by use of a constant. Use case is when the WordPress contents should not be public. Existing public-read default is retained.


philipowen:
Test failure is unrelated to this change so would appreciate a review.

philipowen:
@joehoyle Any thoughts on this please?

joehoyle:
SOrry for the delay @philipowen, this looks good to me!

philipowen:
Fantastic, thanks Joe.

PR #172, opened 2017-05-19
Allow external AWS SDK

### LINKED ISSUE #142: Really use Composer
Please consider using Composer to replace the included aws-sdk in this repo.
https://github.com/humanmade/S3-Uploads/tree/master/lib/aws-sdk

WP.org ready ZIP-s could be uploaded to the releases.

rmccue:
If the AWS SDK is already available, use it. Version compatibility is the responsibility of the user.

Additionally, when loaded via Composer, require the SDK via Composer. This allows Composer to handle version compatibility and such.

Fixes #142.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,6 +19,7 @@
 	},
 	"type": "wordpress-plugin",
 	"require": {
-		"composer/installers": "~1.0"
+		"composer/installers": "~1.0",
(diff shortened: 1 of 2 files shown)


rmccue:
Tests failed, but unrelated.

PR #176, opened 2017-06-30
Improved S3 autoloader performance.

igmoweb:
`s3_uploads_autoload()` was checking if a file exists for any class name passed to it. That was hurting performance in every page load.
The fix tries to recognize if the class is a S3_Uploads family

### DIFF
--- a/s3-uploads.php
+++ b/s3-uploads.php
@@ -97,6 +97,9 @@ function s3_uploads_enabled() {
  * @param $class_name Name of the class to load.
  */
 function s3_uploads_autoload( $class_name ) {
+	if ( strpos( $class_name, 'S3_Uploads' ) !== 0 ) {
+		return;
+	}
 	/*
 	 * Load plugin classes:
 	 * - Class name: S3_Uploads_Image_Editor_Imagick.


igmoweb:
Just another comment I'm curious about. The way `s3_uploads_autoload` loads a class does not match with some classes naming. For instance `S3_Uploads_ChangedFilesIterator` will be parsed to a filename `class-s3-uploads-changedfilesiterator.php` which does not actually exist so it's not this autoloader the one that is loading it. Is it being loaded from another place?

joehoyle:
The idea of this makes sense to me, @rmccue any thoughts?

rmccue:
Seems legit to me. In a modern-standards-world, we'd probably have everything under the `S3_Uploads` namespace instead, which would be a bit of a nicer check.

No idea on the ChangedFilesIterator though; @joehoyle?

archon810:
This is 3 years old and still not merged?

vijayant123:
Does this issue still exist in latest release (version 2.2.1)?

joehoyle:
We now use the "native" composer classmate autoloader, so this is no longer needed. Sorry for how long this sat around for everyone!

Issue #177, opened 2017-06-30
Opening temporary stream ignores any custom S3 putObject configuration


philipowen:
The default case in the anonymous function in `stream_open()` in class-s3-uploads-stream-wrapper.php does not use any custom putObject parameters. This can cause the call to fail.

For example: where a call back is registered against `s3_uploads_putObject_params` that sets encryption key details and the bucket policy says all objects must be encrypted. This _canary_ `putObject()` call doesn't have the encryption details so the call fails.

The solution here seems to be to be to add
`$p = apply_filters( 's3_uploads_putObject_params',  $p );`
On the line before the call to `putObject()`

Perhaps @joehoyle can comment as he wrote that default switch case? I'm happy to submit a PR if it will be useful.

rmccue:
Makes sense to add that filter to me; if you can PR it, happy to merge :)

philipowen:
Thanks. I have created PR #180 with the suggested change in.

PR #180, opened 2017-07-05
add call to apply_filters for test push to S3

philipowen:
Address issue #177 to allow customisation of the call to putObject against S3 that tests the bucket is writable.


rmccue:
Thanks a bunch!

PR #181, opened 2017-07-17
adding section about S3_UPLOADS_DISABLE_REPLACE_UPLOAD_URL to readme.md

jasondewitt:
Felt that this option deserved mention in Readme.md. I was looking for this option and didn't find it in the Readme, had to dig it up from a client site where it is in use. Wanted to add it here if for no other reason than so I can find it the next time I need it.

Feel free to change my wording to be more better.


joehoyle:
Thanks!

PR #185, opened 2017-07-30
Updated the details regarding the region

carl-alberto:
I was trying to setup for the Sydney region and it seems that value is also required so I added a link to the reference page that list the equivalent region for that parameter.


joehoyle:
Thanks!

PR #189, opened 2017-08-12
Fix command case in readme

edcs:
The docs say that you should activate this plugin with the following command:

```
wp plugin activate S3-Uploads
```

It seems like wp-cli is case sensitive and the command should be:

```
wp plugin activate s3-uploads
```

This PR fixes the readme as this threw me for a few minutes while trying to get it up and running!


joehoyle:
This is dependent on what can you clone the repository as, and given this repo name uses `S3-Uploads` it is the default cloned directory name, therefor I think it's a good default for the activate command.

PR #194, opened 2017-08-24
Ignores copy if file path is a directory

dsawardekar:
This PR adds an `is_dir` check alongside the `file_exists` to improve file detection. This fixes warnings in the subsequent `copy` operation where the attachment's `_wp_attached_file` meta field is absent.

```log
Warning: copy(): The first argument to copy() function cannot be a directory in /srv/www/htdocs/wp-content/plugins/s3-uploads/inc/class-s3-uploads-wp-cli-command.php on line 121
```

### DIFF
--- a/inc/class-s3-uploads-wp-cli-command.php
+++ b/inc/class-s3-uploads-wp-cli-command.php
@@ -117,6 +117,9 @@ class S3_Uploads_WP_CLI_Command extends WP_CLI_Command {
 
 		foreach ( $files as $file ) {
 			if ( file_exists( $path = $old_upload_dir['basedir'] . '/' . $file ) ) {
+				if ( is_dir( $path ) ) {
+					continue;
+				}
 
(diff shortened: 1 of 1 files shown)


inc/class-s3-uploads-wp-cli-command.php
```diff
@@ -117,6 +117,9 @@ public function migrate_attachment_to_s3( $args, $args_assoc ) {
 
 		foreach ( $files as $file ) {
 			if ( file_exists( $path = $old_upload_dir['basedir'] . '/' . $file ) ) {
+				if ( is_dir( $path ) ) {
```
joehoyle: I think it would be better to just not add `_wp_attached_file` to the array of files if it's not set, rather than the is_dir check here.

joehoyle:
@dsawardekar thanks for the PR! I think this might be simpler with the attached comment though

joehoyle:
Awesome, thanks!

PR #203, opened 2017-11-14
Update aws sdk to 3.38.1

nathanielks:
Our version of the AWS SDK is very out of date. This brings us up to date. Primary purpose is for the SDK to be able to automatically authenticate IAM Task Roles.


joehoyle:
Thanks!

Issue #184, opened 2017-07-27
How to use it for s3 compitible storage like minio,wasabi,ceph s3, swift?


xianx:
Hello,

How can I use it with s3 compatible storage like minio,wasabi,ceph s3, swift

With awscli there is option of --endpoint-url
http://docs.aws.amazon.com/cli/latest/reference/

How it could be used with S3-Uploads?

etherealite:
This used to work for me when I tried with v2.0.0-beta3
```php
  add_filter('s3_uploads_s3_client_params', function($params){
    error_log('s3 uploads filter run');
    $params['endpoint'] = S3_UPLOADS_BUCKET_URL;
    error_log(var_export($params, true));
    return $params;
  });
```

But that tag is super far behind master, and when I switched to master, this no longer is working for me.

etherealite:
I got it working again using the master branch I think it's on e9775f60afb20ceba1a15d6a73cb58eb8f72f951.

 I used the same hook that lets you edit the config that gets passed to the s3client factory method. The only difference is that I had to add a new parameter, `use_path_style_endpoin` as you can see bellow.
```php
add_filter('s3_uploads_s3_client_params', function($params){
    $params['endpoint'] = 'http://minio-server.dev'    
    $params['use_path_style_endpoint'] = true;¬
    return $params;
  });

joehoyle:
@etherealite thanks for the support help!

PR #208, opened 2017-12-24
Add the ability to define a custom endpoint.

spacedmonkey:
If you are using something like fake s3 or digital ocean spaces, it is useful to be able to define a custom s3 endpoint.

### DIFF
--- a/inc/class-s3-uploads.php
+++ b/inc/class-s3-uploads.php
@@ -172,6 +172,10 @@ class S3_Uploads {
 			$params['region']    = $this->region;
 		}
 
+		if ( defined( 'S3_UPLOADS_ENDPOINT' ) ) {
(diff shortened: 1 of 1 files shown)


joehoyle:
This can already done by existing filters, see https://github.com/humanmade/S3-Uploads/pull/138 and https://github.com/humanmade/S3-Uploads/issues/184. I think I'd prefer to keep it that way rather than more constants. The constants are less than ideal as it makes it much more difficult to have multiple instances of S3_Uploads when needed, as constants are globally set.

roborourke:
@spacedmonkey happy to close this one? Will document how to do this on the readme

spacedmonkey:
@roborourke Works for me. I documented it here as well. https://www.spacedmonkey.com/2018/01/29/using-digital-ocean-spaces-wordpress/

PR #209, opened 2017-12-24
Add the DNS address for the S3 Bucket to list for DNS prefetch.

### LINKED ISSUE #110: Add dns-prefetch link into header
To help performance, you could add a link reference in header to prefetch dns for s3 domain url. In WordPress this would be pretty simple in 4.6 with the new filter. 

``` php

function s3_resource_hints( $hints, $relation_type ) {
    if ( 'dns-prefetch' === $relation_type ) {
        $hints[] = $this->get_s3_url();
    } 

    return $hints;
}

add_filter( 'wp_resource_hints', 's3_resource_hints', 10, 2 );
```

spacedmonkey:
Fixes #110

### DIFF
--- a/inc/class-s3-uploads.php
+++ b/inc/class-s3-uploads.php
@@ -243,7 +243,6 @@ class S3_Uploads {
 	 *
 	 * @param $hints
 	 * @param $relation_type
-	 *
 	 * @return array
 	 */
 	function wp_filter_resource_hints( $hints, $relation_type ) {
--- a/inc/class-s3-uploads.php
+++ b/inc/class-s3-uploads.php
@@ -48,6 +48,7 @@ class S3_Uploads {
 		add_filter( 'wp_image_editors', array( $this, 'filter_editors' ), 9 );
 		add_filter( 'wp_delete_file', array( $this, 'wp_filter_delete_file' ) );
 		add_filter( 'wp_read_image_metadata', array( $this, 'wp_filter_read_image_metadata' ), 10, 2 );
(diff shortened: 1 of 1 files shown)


inc/class-s3-uploads.php
```diff
@@ -237,6 +238,22 @@ public function wp_filter_read_image_metadata( $meta, $file ) {
 		return $meta;
 	}
 
+	/**
+	 * Add the DNS address for the S3 Bucket to list for DNS prefetch.
+	 *
+	 * @param $hints
+	 * @param $relation_type
+	 *
```
peterwilsoncc: CS: Remove empty line above `@return`.

inc/class-s3-uploads.php
```diff
@@ -237,6 +238,22 @@ public function wp_filter_read_image_metadata( $meta, $file ) {
...
+	 * @return array
+	 */
+	function wp_filter_resource_hints( $hints, $relation_type ) {
+		if ( 'dns-prefetch' === $relation_type ) {
```
peterwilsoncc: This could use the `if ( !== ) { return $hints }` pattern but I think it's clearer and less error prone this way.

Values and names checked, all correct.

spacedmonkey: I did think about doing it this way, but personally I thought this way was easier to read. Willing to change it if anyone disagrees...

spacedmonkey:
@peterwilsoncc Requested change has been done in 4920d934c8abeae90960f3cfff10fa849dc10f91 :+1:

spacedmonkey:
@peterwilsoncc @joehoyle Any chance of a merge for this one? I think I done all the feedback as requested.

Issue #80, opened 2015-12-18
plugin collisions when other plugin uses guzzle


ivoba:
hi,
i just experienced an error similar to this https://github.com/humanmade/S3-Uploads/issues/24 

```
'Guzzle\Common\Exception\RuntimeException' with message 'Guzzle\Http\CachingEntityBody supports only SEEK_SET and SEEK_CUR seek operation
```

although i was already on the latest version where the hack for guzzle was applied:  
https://github.com/humanmade/S3-Uploads/commit/a9db1096ca5a042386073851c96061102265b309

it turned out i was also using sendgrid-email-delivery-simplified plugin and a composer loaded guzzle which was somehow replacing your shipped guzzle and the hack had no effect anymore.

just to let you know.

joehoyle:
Ahh, good point - the AWS SDK hacks we have is super nasty, so I'd really like to remove them - let's keep this open as a reminder of the downsides.

joshuajabbour:
Bummer, these hacks are making it so we can't use this plugin...

joehoyle:
All AWS SDK hacks have been removed. It's also recommended to installed S3 Uploads using composer, so all your plugins can share Guzzle instances.

PR #212, opened 2017-12-30
Use composer for AWS lib

davisshaver:
Also fixes https://github.com/humanmade/S3-Uploads/issues/80


joehoyle:
Dupe of https://github.com/humanmade/S3-Uploads/pull/101, but thanks for the PR!

PR #214, opened 2018-01-12
Multisite support

JoelBernerman:
Made the plugin use site options if mutlisite is activated.
Now rewrites all the urls.

Also removed the Aws\S3\Sync\UploadSyncBuilder extend because it does not seem to be compatible with current AWS SDK included.
This might break the sync option when you run upload-directory in --sync mode but at least it works with the default upload-directory.

### DIFF
--- a/inc/class-s3-uploads-uploadsyncbuilder.php
+++ b/inc/class-s3-uploads-uploadsyncbuilder.php
@@ -1,6 +1,6 @@
 <?php
 
-class S3_Uploads_UploadSyncBuilder extends Aws\S3\Sync\UploadSyncBuilder {
+class S3_Uploads_UploadSyncBuilder extends {
 
 	public function __construct( $is_dry_run = false ) {
 		$this->dry_run = $is_dry_run;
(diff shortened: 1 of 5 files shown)


spacedmonkey:
@JoelBernerman Just a heads up `update_site_option` if you use it in single site, it falls back to standard options. :+1:

joehoyle:
This will change existing bahavour so I think doesn't make to merge in this state. Right now you are _meant_ to be able to have it enabled on a per-site basis, from what I'm reading, this will do away with that. @JoelBernerman could you explain why you'd want to toggle it at the network level more than the site level? I think we'd need to come up with some mechanism to support both

spacedmonkey:
You can network enable S3, using a filter in an MU plugin 
```
add_filter('pre_option_s3_uploads_enabled, function($current){
   return 'enabled';
});
```

JoelBernerman:
Well you are right.
It will break the per site enable possibility.
Did not really think about that use case.
We use multisite without any domain mapping so all blogs are quite strongly connected to the main site.
Like bloggers to our brand sites or event-specific subdomains for the brands.

Ill try out the filter-solution.

JoelBernerman:
To further explain why I wanted this, it would just be a big hassle to manually enable something for every blog on our sites.
We can have 50+ blogs in our network and using the plugin "as is" it would require us to manually enable every one.

Issue #92, opened 2016-04-13
Permanently deleted media files


larssn:
Hey, good plugin, works well.

Quick question that may or may not be an issue:
When deleting a media file, the originally uploaded file gets left behind - only the different thumbnail variants gets deleted. Is this intended behavior?

We run a multisite setup, and would prefer that all files permanently deleted through Wordpress would get removed completely from S3.

Thanks.

Ps. I can delete the left over file through the CLI, so at least that is working, although not convenient:
wp s3-uploads rm uploads/2016/04/image.jpg --debug --url=www.our-multisite-setup.com

larssn:
Further strangeness: I remove the wp_delete_file filter from S3_Uploads: Then the original file gets deleted, but the thumbnail variants doesn't. I can understand why they dont get deleted, but now I don't understand why the original file DOES get deleted... 🎱

joehoyle:
Oops think I broke this in https://github.com/humanmade/S3-Uploads/pull/87 whereby I fixed the removing of sizes, but that apparently broke the original

larssn:
Where does the delete hook occur? Been looking through the code, and all I see is the rm function in the CLI class (which seems irrelevant), and the wp delete filter that you added.

joehoyle:
WordPress deletes the file internally, and just passes it through the `wp_delete_file` hook first, which I tinkered with. Will get a patch up for this.

larssn:
👍
Thanks for the help!

larssn:
Forgive me, I'm new when it comes to navigating github. Is there a section for patches? Been looking around for it.

thefrosty:
Will this get patched in `1.1.1` or in the `2.x` release?

nuvoPoint:
We are having the same issue. Is there a fix for this or should we make a fork?

josephabrahams:
:+1:

mathewemoore:
Is anyone working on a fix for this? I am working on a project and everything is working fine except for deleting the original file.

MordiSacks:
I went through the whole project, for the life of me I can't understand why main file does not get deleted.

ikappas:
Hi all,

When an attachment is deleted the function `wp_delete_attachment` is called that checks whether the attachment has thumbnail, intermmediate and backup images and removes them with the following code:

```php
$thumbfile = apply_filters( 'wp_delete_file', $thumbfile );
@ unlink( path_join($uploadpath['basedir'], $thumbfile) );
```
[...]

PR #215, opened 2018-01-12
Don't filter absolute paths in wp_delete_file()

jeremyfelt:
The `wp_delete_file` filter  is used in two different ways by WordPress: 

1. When filtering the filename for the removal of thumbnails, extra image sizes, and backup sizes in `wp_delete_attachment()`, a relative path is expected in return. This relative path is then used to rebuild the absolute path before `unlink()` fires.
1. When filtering directly in `wp_delete_file()`, an absolute path is expected at all times.

We can identify when each of these is used by capturing the original file's absolute path when `delete_attachment` fires and only leaving that untouched. When `wp_delete_file()` is used directly, no original file is captured, so the path is left untouched.

This fixes a case where original files would not be deleted from S3 when removed from the media library, but where all of the thumbnails would be deleted.

See https://github.com/humanmade/S3-Uploads/issues/92
See https://core.trac.wordpress.org/ticket/39476

### DIFF
--- a/inc/class-s3-uploads.php
+++ b/inc/class-s3-uploads.php
@@ -9,6 +9,7 @@ class S3_Uploads {
 	private        $secret;
 
 	public $original_upload_dir;
+	public $original_file;
 
 	/**
 	 *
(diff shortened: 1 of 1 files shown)


joehoyle:
This looks good to me, any chance of a unit test for this case @jeremyfelt?

PR #219, opened 2018-01-29
Simple CLoudFront support

wieshka:
This adds simple yet effective ability to use CloudFront distributed assets rather S3 web.


joehoyle:
It's recommended to just use the `s3_uploads_bucket_url` in this case, rather than supporting yet another (globally scoped) constant.

wieshka:
It was faster to hack then read the doc, found S3_UPLOADS_BUCKET_URL, canceling.

PR #232, opened 2018-04-10
fix upload ContentType guess based on the key object file extension

zvfvrv:
wordpress append the .tmp extension to filenames when generating thumbnails. E.g.
<file-name>-150x150.jpg.tmp
So when guessing the ContentType from the extension the Psr7\mimetype_from_filename() will not work properly.
To avoid this issue, we strip the .tmp extension if present.

### DIFF
--- a/inc/class-s3-uploads-stream-wrapper.php
+++ b/inc/class-s3-uploads-stream-wrapper.php
@@ -188,15 +188,15 @@ class S3_Uploads_Stream_Wrapper
 		$params['Body'] = $this->body;
 
 		//Added by EveryUP Srl
-		$objectKey = $params['Key'];
-		if(pathinfo($objectKey, PATHINFO_EXTENSION) == 'tmp') {
-			$objectKey = substr($objectKey, 0, -4);
+		$object_key = $params['Key'];
+		if ( pathinfo($object_key, PATHINFO_EXTENSION) === 'tmp' ) {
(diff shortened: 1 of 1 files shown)


joehoyle:
@zvfvrv could you point me to where in WordPress Core the `tmp` is applied? Thanks!

zvfvrv:
Hi @joehoyle , 
unfortunately i dunno, i realized that for each thumbnails automatically generated from wordpress,  the value of `$params['Key']` is like  filename-150x150.jpg.tmp or filename-600x486.png.tmp.
Regards.

joehoyle:
Hmm ok, seems like we need to investigate if this behavior is from Core or potentially a plugin, etc.

zvfvrv:
@joehoyle i have investigated on a wp instance with the issue, perhaps the problem is related to a plugin for image optimization WP SMUSH.

wp-networks:
@joehoyle for a start, perhaps see [wp-admin/includes/file.php L610](https://core.trac.wordpress.org/browser/branches/4.9/src/wp-admin/includes/file.php#L610) (and [wp-smushit/lib/class-wpsmush.php L248](https://plugins.trac.wordpress.org/browser/wp-smushit/trunk/lib/class-wp-smush.php#L248))

joehoyle:
Thi is very stale now due to how old it is, and we use the updaed SDK so I'm not sure if this is still an issue.

PR #234, opened 2018-04-25
Support for custom base path for uploads

ViliamKopecky:


pySilver:
@ViliamKopecky I assume this CI failure has nothing to do with actual feature?

devbhosale:
@ViliamKopecky  this is a useful feature but merge is failing.. can you help me understand where else changes are required and if I can help complete this in any way if you don't have the time?

ViliamKopecky:
I'm not sure what to do. Seems to me, that the test errors are not caused by any of my changes... ?

devbhosale:
@joehoyle  @rmccue  can one of you help please?

joehoyle:
@devbhosale I'm not sure this is a common enough case to add a constant for. Is it possible to use the existing hooks to do this via an addon? If not, it might be a better approach to add whatever filters we would need to do that. Does that make sense?

devbhosale:
@joehoyle the PR is about using a directory and sub-directories on an S3 bucket. Don't you think it is kind of basic functionality to organize files into folders? Even if a client uses a single bucket, they might still want to use directories. I thought @ViliamKopecky built a clean solution..  

However, if you still disagree. Feel free to elaborate on your ideas for a filter.

joehoyle:
@devbhosale you have already do that via `S3_UPLOADS_BUCKET` by setting it to `my-bucket/my-path

ViliamKopecky:
@joehoyle You sure it works? I think there are places where it actually expects the bucket name, which i suppose should not be valid bucket name with some path after it. But i should try it, maybe AWS takes it well.

Some filters would be probably better in the wordpress way.

joehoyle:
@ViliamKopecky absolutely, we use this method on virtually every site we use S3 Uploads on, as we have a shared bucket. This is why we do `strtok` here for example: https://github.com/humanmade/S3-Uploads/blob/master/inc/class-s3-uploads.php#L156

devbhosale:
@joehoyle  @ViliamKopecky  tested it .. suffixing directory path on bucket works... 

this should have "doc update" PR request from the beginning... saving everyone's time...  :-)  I will create one right now..

joehoyle:
@devbhosale yeah, very sorry about that not being clear at all!

joehoyle:
Closing this one out either way I think

PR #238, opened 2018-05-15
Include updates getID3 lib

joehoyle:
This is needed as getID3 lib that's bundled with Core is out of data and doesn't work with s3:// paths. See https://core.trac.wordpress.org/ticket/43836

We'll likely want to remove this one day once core supports a newer version, however this isn't likely to happen soon as the latest released vesion of getid3 still doesn't encorperate the fix needed.


joehoyle:
Explanation comment at https://github.com/humanmade/S3-Uploads/pull/238/commits/c6d87002e569b6a24a22735fec55da3471a15384#diff-71b5946c9cf03cbcd5451dc879ed1ddbR47 and tests at https://github.com/humanmade/S3-Uploads/pull/238/commits/c6d87002e569b6a24a22735fec55da3471a15384#diff-5b98629e5b12646b1db8b196fa4ac1eeR97

roborourke:
Merging, there is an issue with HM Linter hitting its memory limit but I don't believe anything in that lib will meet our coding standards so I think we need to add a phpcs ruleset to exclude it.

rmccue:
@roborourke This should have been fixed before merging, not after.

PR #243, opened 2018-05-28
Override the Personal Data Export directory

joehoyle:
Add filters to "wrap" the wp_privacy_personal_data_export_file function call as we need to  switch out the personal_data directory to a local temp folder, and then upload after it's complete, as Core tries to write directly to the ZipArchive which won't work with the S3 streamWrapper.

### DIFF
--- a/s3-uploads.php
+++ b/s3-uploads.php
@@ -14,6 +14,14 @@ if ( defined( 'WP_CLI' ) && WP_CLI ) {
 
 add_action( 'plugins_loaded', 's3_uploads_init' );
 
+// Add filters to "wrap" the wp_privacy_personal_data_export_file function call as we need to
+// switch out the personal_data directory to a local temp folder, and then upload after it's
+// complete, as Core tries to write directly to the ZipArchive which won't work with the
+// S3 streamWrapper.
+add_action( 'wp_privacy_personal_data_export_file', 's3_uploads_before_export_personal_data', 9 );
+add_action( 'wp_privacy_personal_data_export_file', 's3_uploads_after_export_personal_data', 11 );
+add_action( 'wp_privacy_personal_data_export_file_created', 's3_uplodas_move_temp_peronal_data_to_s3' );
+
 function s3_uploads_init() {
 	// Ensure the AWS SDK can be loaded.
 	if ( ! class_exists( '\\Aws\\S3\\S3Client' ) ) {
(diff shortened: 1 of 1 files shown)


s3-uploads.php
```diff
@@ -14,6 +14,14 @@
 
 add_action( 'plugins_loaded', 's3_uploads_init' );
 
+// Add filters to "wrap" the wp_privacy_personal_data_export_file function call as we need to
+// switch out the personal_data directory to a local temp folder, and then upload after it's
+// complete, as Core tries to write directly to the ZipArchive which won't work with the
+// S3 streamWrapper.
+add_action( 'wp_privacy_personal_data_export_file', 's3_uploads_before_export_personal_data', 9 );
+add_action( 'wp_privacy_personal_data_export_file', 's3_uploads_after_export_personal_data', 11 );
+add_action( 'wp_privacy_personal_data_export_file_created', 's3_uplodas_move_temp_personal_data_to_s3', 1000 );
```
rmccue: Can we hook these in `s3_uploads_init()` instead? Right now, they'll be hooked in even if S3 Uploads isn't active (e.g. missing bucket constant).

s3-uploads.php
```diff
@@ -122,3 +130,57 @@ function s3_uploads_autoload( $class_name ) {
...
+ * @return string
+ */
+function s3_uploads_set_wp_privacy_exports_dir( $dir ) {
+	if ( strpos( $dir, 's3://' ) !== 0 ) {
+		return $dir;
+	}
+	$dir = sys_get_temp_dir() . '/wp_privacy_exports_dir/';
+	if ( ! is_dir( $dir ) ) {
+		mkdir( $dir );
+		file_put_contents( '', $dir . 'index.html' );
```
rmccue: Your parameters are backwards.

s3-uploads.php
```diff
@@ -122,3 +130,57 @@ function s3_uploads_autoload( $class_name ) {
...
+ * location to the S3 location where it should have been stored all along, and where
+ * the "natural" Core URL is going to be pointing to.
+ */
+function s3_uplodas_move_temp_personal_data_to_s3( $archive_pathname ) {
+	if ( strpos( $archive_pathname, sys_get_temp_dir() ) !== 0 ) {
+		return;
+	}
+	$upload_dir  = wp_upload_dir();
+	$exports_dir = trailingslashit( $upload_dir['basedir'] ) . 'wp-personal-data-exports/';
+	copy( $archive_pathname, $exports_dir . pathinfo( $archive_pathname, PATHINFO_FILENAME ) . '.' . pathinfo( $archive_pathname, PATHINFO_EXTENSION ) );
```
rmccue: Can we break the filename building on to a separate line here? It's very long otherwise.

s3-uploads.php
```diff
@@ -52,6 +52,14 @@ function s3_uploads_init() {
...
 		require_once dirname( __FILE__ ) . '/lib/getid3/getid3.php';
 	}
+
+	// Add filters to "wrap" the wp_privacy_personal_data_export_file function call as we need to
+	// switch out the personal_data directory to a local temp folder, and then upload after it's
+	// complete, as Core tries to write directly to the ZipArchive which won't work with the
+	// S3 streamWrapper.
+	add_action( 'wp_privacy_personal_data_export_file', 's3_uploads_before_export_personal_data', 9 );
+	add_action( 'wp_privacy_personal_data_export_file', 's3_uploads_after_export_personal_data', 11 );
+	add_action( 'wp_privacy_personal_data_export_file_created', 's3_uplodas_move_temp_personal_data_to_s3', 1000 );
```
rmccue: You've misspelled the function here and below.

PR #244, opened 2018-05-30
Update for aws-sdk-php v3

dingo-d:
Removed unused methods, and updated some to use v3 of sdk, removed lib from the plugin and used composer to autoload and install latest version of aws-sdk-php.

The plugin should be manually bundled once it's tested and then put in a release as a zip. Unit tests failed locally because you need to provide S3 details for it to work.


joehoyle:
We now pull in the aws-sdk via composer in https://github.com/humanmade/S3-Uploads/blob/master/composer.json#L23, but thanks for the PR @dingo-d sorry this slipped through the net

Issue #246, opened 2018-06-08
Conflict with caching plugin (Comet Cache Pro)


nathanegraham:
This is a great plugin. Thanks for your work on it. I've not seen this error before but I'm getting the below fatal error when activating on a site with Comet Cache Pro enabled:

`Fatal error: Cannot redeclare GuzzleHttp\uri_template() (previously declared in /www/eng/coursedev-ep/wp-content/plugins/comet-cache-pro/src/vendor/guzzlehttp/guzzle/src/functions.php:17) in /www/eng/wp-content/plugins/S3-Uploads/lib/aws-sdk/GuzzleHttp/functions.php on line 31`

PR #253, opened 2018-08-02
issue-246-fix-plugin-guzzle-function-dependency-collision: 1) fix the…

nightelf:
Fixes issue 246: https://github.com/humanmade/S3-Uploads/issues/246
1) Fixes the problem of guzzle functions inclusion collisions when more than 1 plugin uses guzzlehttp as a dependency. It checks for the existence of the first function in the php file before loading it.
2) change require to require_once. No need to load these files more than once.

### DIFF
--- a/lib/aws-sdk/aws-autoloader.php
+++ b/lib/aws-sdk/aws-autoloader.php
@@ -754,12 +754,23 @@ $mapping = array(
 
 spl_autoload_register(function ($class) use ($mapping) {
     if (isset($mapping[$class])) {
-        require $mapping[$class];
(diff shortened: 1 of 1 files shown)


dienncrelipa:
Merge ,please

doquangthanh:
I have another way to fix this problem is to use the composer autoload please see the code here https://github.com/doquangthanh/S3-Uploads

joehoyle:
S3 Uploads is now "properly" using the composer autoloader and dev system generally. That should mean if you are installing with composer, this should no longer be an issue.

Issue #257, opened 2018-08-31
 Autoload GuzzleHttp's functions_include.php files instead of functions.php


takkaria:
The AWS SDK [autoloads three functions.php files from GuzzleHttp](https://github.com/humanmade/S3-Uploads/blob/master/lib/aws-sdk/aws-autoloader.php):

```
require __DIR__ . '/GuzzleHttp/functions.php';
require __DIR__ . '/GuzzleHttp/Psr7/functions.php';
require __DIR__ . '/GuzzleHttp/Promise/functions.php';
```

When you have multiple vendored versions of the GuzzleHttp library on your site, this causes clashes because you get function redefinition fatal errors.  This is similar to what happened with #129, except I don't have the luxury of using dependency management to get around it.

GuzzleHttp does provide `functions_include.php` files, which have include guards in them to prevent redefinition errors, but neither Guzzle nor AWS currently use them.  (I've filed bugs upstream about this: for [Guzzle](https://github.com/guzzle/guzzle/issues/2135) and for [AWS](https://github.com/aws/aws-sdk-php/issues/1616).)  In the meantime, it would be great if your vendored version of the AWS SDK could use them.

asretux:
We also need this change to fully automate an ongoing project's dependencies.

PR #258, opened 2018-08-31
Use include-guarded functions.php files in AWS SDK

takkaria:
See #257


takkaria:
Hi, just wondering if anyone had a chance to look at this yet?  It's a very simple fix, and the upstream AWS SDK has made the [same change](https://github.com/aws/aws-sdk-php/pull/1624/files).

joehoyle:
Sorry for never picking this up @takkaria. We now load the SDK via composer so this isn't relevant anymore.

PR #259, opened 2018-08-31
Constant to define custom hook priority for URL Rewriting

viniciusbarros:
Hi,
I noticed an incompatibility when using this plugin while developing for an existent project that already does some URL rewriting. I am pretty sure that this incompatibility was happening because the setup we have uses [bedrock](https://roots.io/bedrock/)

So I added:
- A new constant to define default WP hook priority
- A new constant setting `S3_UPLOADS_RENAME_FOLDER_PRIORITY ` that developers can use to define a custom URL Rewrite priority if needed.

I tested it locally and it works fine. In my case I had to bump the permission to 11, instead of 10.
Let me know if this can be added or if something else needs to be amended.

Thank you!
Vinicius

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,7 +153,10 @@ S3 Uploads' URL rewriting feature can be disabled if the current website does no
 // disable URL rewriting alltogether
 define( 'S3_UPLOADS_DISABLE_REPLACE_UPLOAD_URL', true );
 ```
-
+Sometimes due to conflicts with other plugins, you may want to have the URL Rewrite functionality to be executed using a different hook priority. For that you can specify it by setting the constant `S3_UPLOADS_RENAME_FOLDER_PRIORITY` in your configurations file. The default priority value is 10.
+```
+define('S3_UPLOADS_RENAME_FOLDER_PRIORITY', 10);
+```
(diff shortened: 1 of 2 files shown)


viniciusbarros:
Hello, I think the automated build check was failing previous to this Pull Request.

viniciusbarros:
Hi guys, any estimate for this PR to be reviewed?
I really would like to be able to run the official plugin w/ bedrock setup.

joehoyle:
Hmm I'm not really keen on this approach, I feel like a constant here is a bit of a wild edgecase and the configuration this way is a bit too much messing with internals.

I'd rather support this via pulling custom coding manually just registering the hooks its self, rather than supporting it here. Thinking something like you could just had an addon plugin that did:

```php
remove_filter( 'upload_dir', array( S3_Uploads::get_instance(), 'filter_upload_dir' );
add_filter( 'upload_dir', array( S3_Uploads::get_instance(), 'filter_upload_dir', 2 );
...
```

Or whatever you might want.

joehoyle:
Closing this, and suggestion is to use the above workaround if you really need to change the priority.

PR #262, opened 2018-09-11
Trigger putObject action on local wrapper as well

shadyvb:
This facilitates local development by firing the same action that's fired on live environment as well.


shadyvb:
HM Linter is failing to run because of memory issues, and Travis is failing on errors unrelated to the changes ( same failures as on `master` ).

PR #266, opened 2018-09-15
Fix ACL on files uploaded via upload-directory

bosunolanrewaju:
This fixes ACL permission on files uploaded through `upload-directory` command. 

```
wp s3-uploads upload-directory wp-content/uploads s3://uploads
```

Files in the upload directory uploaded through command above are not accessible to the public despite the ACL `public-read` permission set by default.

Going by the fix to the issue [here](https://github.com/aws/aws-sdk-php/issues/867#issuecomment-169147302). I was able to set the right permissions for all the files in the directory.

### DIFF
--- a/inc/class-s3-uploads-wp-cli-command.php
+++ b/inc/class-s3-uploads-wp-cli-command.php
@@ -324,8 +324,8 @@ class S3_Uploads_WP_CLI_Command extends WP_CLI_Command {
 				$prefix . $to,
 				array(
 					'debug'       => true,
-					'params'      => function(\AWS\Command $command) {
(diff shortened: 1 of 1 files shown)


bosunolanrewaju:
@emanuelpoletto Thanks for pointing out the attribute name. And checking for constant definition is a good move. I will update the PR. 

Thank you for sharing.

joehoyle:
This was fixed in https://github.com/humanmade/S3-Uploads/commit/d75ca22a55d73847918ead42893b88162689bc25 Thanks for the PR @bosunolanrewaju

PR #268, opened 2018-09-24
Add custom endpoint

GasimGasimzada:
This commit adds `S3_UPLOADS_ENDPOINT` constant and allows the developer to change the endpoint if using custom S3 compatible object storages. I have tested it with the service that I am using currently and all the operations work as expected.

Additionally, I think it would be beneficial for developers to know which services the plugin is tested with in the README or wiki.

### DIFF
--- a/inc/class-s3-uploads.php
+++ b/inc/class-s3-uploads.php
@@ -7,6 +7,7 @@ class S3_Uploads {
 	private        $bucket_url;
 	private        $key;
 	private        $secret;
+	private 	   $endpoint;
 
 	public $original_upload_dir;
 	public $original_file;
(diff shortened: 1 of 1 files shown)


spacedmonkey:
@GasimGasimzada This can be done in your own plugin. See my blog post here. https://www.spacedmonkey.com/2018/01/29/using-digital-ocean-spaces-wordpress/

rmccue:
Closing, as this is possible in other plugins. As we do not use non-S3 endpoints at Human Made, [we won't offer official support for non-S3 hosts](https://engineering.hmn.md/projects/support/), but the filter @spacedmonkey mentioned provides this capability.

PR #270, opened 2018-10-11
Update docs for bucket path prefix

devbhosale:
Update docs for bucket path prefix and include an example to show how to use folders


rmccue:
Thanks!

Issue #283, opened 2018-12-19
Tag a new version ⚠️


tristanbes:
Hello,

Last tagged version is v2 in **march**, there is critical FIXES which has been made and not available in v2 (for example https://github.com/humanmade/S3-Uploads/pull/255)

Can you tag a new version so I don't have to require `dev-master` branch using composer ? Thanks.

tristanbes:
Please, can we get the tag already ? (and not a RC) :)

Kocal:
It would be so nice to have a new tagged release, at least for #255 that is really cumbersome... :confused:

Thanks!

tristanbes:
ping @joehoyle // @rmccue

joehoyle:
Waiting on https://github.com/humanmade/S3-Uploads/pull/293 to go with version 2.

joehoyle:
This is now done.

Kocal:
Thanks!

PR #284, opened 2018-12-19
Fix installation procedure

tristanbes:
when typing `S3-Uploads` it says no plugin found.
when typing `s3-uploads` it correctly activates the plugin

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ define( 'S3_UPLOADS_BUCKET', 'my-bucket/my-folder' );
 You must then enable the plugin. To do this via WP-CLI use command:
 
 ```
-wp plugin activate S3-Uploads
+wp plugin activate s3-uploads
 ```
 
 The next thing that you should do is to verify your setup. You can do this using the `verify` command


rmccue:
Thanks for your PR!

The name here needs to match whatever the directory name for the plugin is. As this repo is called `S3-Uploads`, most people will clone it down with that name, so the instructions here are already correct as far as I can see. They are also only indicative, as it's just telling you how to activate a plugin.

I'm going to close this for now, but feel free to comment if you think we can solve this better.

tristanbes:
Well `composer req humanmade/s3-uploads` clones as `s3-uploads`

Maybe add an alternative line something like: 

If you downloaded the plugin with composer, run `wp plugin activate s3-uploads`

Even if it's just information, nothing is more irritative than copy/paste commands that doesn't work.

tristanbes:
And also @rmccue  please consider releasing a new version via github tags, you got users that use your plugin with composer that is stuck with a very old version, not all of them download via wordpress. See: https://github.com/humanmade/S3-Uploads/issues/283

rmccue:
Our priority for our code is always our internal projects and engineers. We make these projects available as open source as a way of giving back to the community (which also helps us out with contributions), but we can't support things we're not using without doing work for free. :)

tristanbes:
See this with another eye: tagging releases with github respecting the `SemVer` constraint has nothing to do with composer. :-)

It's just a good practice when you do something open source.

rmccue:
Sure, which is why we'll get to it eventually, but internally we use submodules tied to commits, so tags are mostly irrelevant. :)

PR #305, opened 2019-04-08
Remove composer deps

joehoyle:
I think for v2 we should remove the composer deps from the master branch, and tagged version should be from that branch. THis means installing the plugin via composer won't cause additoinal copies of the AWS SDK to exist.

We can still have a build branch, and upload artifacts to the release page I believe, for people who want to use the plugin without composer.


rmccue:
@joehoyle did you remove the hacks to the AWS SDK? I thought we still had some, which necessitated not using Composer/etc here.

joehoyle:
@rmccue that was a long time ago, those haven't been needed for a while

s3-uploads.php
```diff
@@ -18,7 +18,7 @@ function s3_uploads_init() {
 	// Ensure the AWS SDK can be loaded.
 	if ( ! class_exists( '\\Aws\\S3\\S3Client' ) ) {
```
rmccue: I feel like we should maybe change this to instead be `if file_exists vendor/autoload` instead now?

joehoyle: Hmm I think it's better the way it already was - including the autoload file will cause issues is `S3Client` is already loaded via another autoload (I think).

tylerlindell: this is causing issues as it thows a fatal error stating `No such file or directory in /wp-content/plugins/S3-Uploads-2.2.1/vendor/autoload.php`.

rmccue: @tylerlindell You need to use Composer to install the dependencies when updating. (Also, please file a new issue rather than commenting on old PRs, otherwise we can easily miss it!)

PR #306, opened 2019-04-08
Fix IMagick with s3 compat

joehoyle:
In some cases (it's not that clear how) Imagick is not compatible with remote streams. According to the docs, this has never been possible, but I think it happened to work in the past. I've tried the same version of imagick, on different sets and sometimes works, some times doesnt. The crux of the issue is that IMagick tries to read from the stream, and is sending a SEEK request to something near PHP_MAX_INT, which causes memory allocation issues on our stream wrapper. It's maybe possible to work around it with a seek stream guard, but really I think that's just one of many possible issues. I think it's going to be better to let IMagick always work with local images to avoid issues.

### DIFF
--- a/inc/class-s3-uploads-image-editor-imagick.php
+++ b/inc/class-s3-uploads-image-editor-imagick.php
@@ -5,6 +5,48 @@ class S3_Uploads_Image_Editor_Imagick extends WP_Image_Editor_Imagick {
 	protected $temp_file_to_cleanup = null;
 
 	/**
+	 * Hold on to a reference of all temp local files.
+	 *
+	 * These are cleaned up on __destruct.
+	 *
+	 * @var array
+	 */
+	protected $temp_files_to_cleanup = array();
+
+	/**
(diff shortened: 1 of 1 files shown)


inc/class-s3-uploads-image-editor-imagick.php
```diff
@@ -47,40 +89,8 @@ protected function _save( $image, $filename = null, $mime_type = null ) {
...
-
-	/**
-	 * Sets up Imagick for PDF processing.
-	 * Increases rendering DPI and only loads first page.
-	 *
-	 * @since 4.7.0
-	 *
-	 * @return string|WP_Error File to load or WP_Error on failure.
-	 */
-	protected function pdf_setup() {
```
roborourke: Why remove this?

joehoyle: PDF handling is now called in `load`, and given now we are preparing the the `load` with a local disk file, this workaround is not longer needed.

PR #307, opened 2019-04-10
Fix delete files on delete attachment

joehoyle:
In WordPress 4.9.7, attachment file deleting changed again, which mangles the s3:// paths very badly. What's more, there's no hooks to be able to adjust / change this behaviour. More details in the code comment.

### DIFF
--- a/inc/class-s3-uploads.php
+++ b/inc/class-s3-uploads.php
@@ -47,8 +47,7 @@ class S3_Uploads {
 
 		add_filter( 'upload_dir', array( $this, 'filter_upload_dir' ) );
 		add_filter( 'wp_image_editors', array( $this, 'filter_editors' ), 9 );
-		add_action( 'delete_attachment', array( $this, 'set_original_file' ) );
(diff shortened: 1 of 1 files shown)


inc/class-s3-uploads.php
```diff
@@ -106,35 +104,29 @@ public function filter_upload_dir( $dirs ) {
...
+	 * WordPress Core's handling of deleting files for attachments via
+	 * wp_delete_attachment_files is not compatible with remote streams, as
+	 * it makes many assumptions about local file paths. The hooks also do
+	 * not exist to be able to modify their behavior. As such, we just clean
+	 * up the s3 files when an attachment is removed, and leave WordPress to try
+	 * a failed attempt at mangling the s3:// urls.
 	 *
-	 * @param  string $file_path
-	 * @return string
+	 * @param $post_id
```
roborourke:
```suggestion
	 * @param int $post_id
```

johnbillion:
@roborourke Yeah this was a fix for a core security issue. See https://codex.wordpress.org/Version_4.9.7.

roborourke:
@johnbillion ah cool, thanks for the context :)

PR #339, opened 2019-09-02
Don't add resource hints if no external S3 URL is used

ocean90:
If `S3_UPLOADS_DISABLE_REPLACE_UPLOAD_URL` or `S3_UPLOADS_USE_LOCAL` is set, the prefetch resource hint shouldn't be registered to prevent an unused look up.


joehoyle:
Thanks @ocean90!

PR #346, opened 2019-10-01
Using wp_delete_file() instead of unlink()

luisfavila:
The use of unlink() breaks functionality of some third-party WordPress plugins.
The use of wp_delete_file() is preferred over unlink() since it allows filtering the file before deletion
Please consider including this in your codebase.


joehoyle:
Thanks @luisfavila !

Issue #327, opened 2019-06-11
Unknown --dry-run and --sync parameters with upload-directory command


kodie:
Running `wp s3-uploads upload-directory content/uploads --sync` or `wp s3-uploads upload-directory content/uploads --dry-run` on the latest plugin version results in

```
Error: Parameter errors:
 unknown --sync parameter
```

and

```
Error: Parameter errors:
 unknown --dry-run parameter
```

jemediacorp:
Receiving the exact same error after installing the plugin for the first time today from a cloned copy of the Git repository (complete with running composer install as the documentation suggests). Here's the plugin info reported by wp-cli:


Plugin S3-Uploads details:
    Name: S3 Uploads
    Status: Active
    Version: 2.1.0-RC2
    Author: Human Made Limited
    Description: Store uploads in S3

PR #351, opened 2019-10-11
Remove upload-directory --sync and --dry-run cli options from documentation as they were removed in fef7efb

kodie:
It seems like the `--sync` and `--dry-run` cli options were removed in fef7efb but were not removed from documentation which caused some confusion and I opened issue #327. This pull request simply removes them from README.md.


joehoyle:
Thank you @kodie

PR #352, opened 2019-10-18
Update file comment header to reflect latest version

theasteve:
Update comment version from 2.1.0-RC2 to 2.1.0.


joehoyle:
Fixed directly, but thanks for the PR @theasteve!

PR #354, opened 2019-10-22
Update to newer version of AWS SDK call for IAM

bradydowling:
This fixes https://github.com/humanmade/S3-Uploads/issues/95 and implements the solution proposed by @dingo-d

Please have a close look at this as I am simply copy/pasting in the code he referenced there, which worked for me as well.


joehoyle:
In v3 of s3-uploads we have removed the ability to create the IAM user. Instead you can still generate the policy, and either manually create the IAM user, or provision it with terraform/cloudformation. 

Sorry for letting this sit @bradydowling

PR #340, opened 2019-09-02
Create a temporary file and use it during verification via CLI

ocean90:
Since the test folder is no longer included in production builds the verification currently fails due to the missing image. This PR replaces the use of the image with a temporary file.

Fixes #309.
Fixes #332.

PR #357, opened 2019-11-15
Fix verify WP CLI command

nathanielks:
`tests/` isn't included in the release package, so this instead uses a small text file to use for verification instead.

### DIFF
--- a/verify.txt
+++ b/verify.txt
@@ -1,0 +1,1 @@
+Beep boop
--- a/verify.txt
+++ b/verify.txt
@@ -1,1 +1,1 @@
-Beep boop
+• Todo: migrate to Altis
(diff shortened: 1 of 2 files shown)


verify.txt
```diff
@@ -0,0 +1 @@
+Beep boop
```
rmccue: I'd also have accepted "• Todo: migrate to Altis"

nathanielks: Updated 😂

ocean90:
Any reason why https://github.com/humanmade/S3-Uploads/pull/340 wasn't used instead of this PR?

nathanielks:
Hey @ocean90! No, not intentionally. I didn't see your PR when I made this one. I appreciate your contribution!

PR #370, opened 2019-12-22
Update getID3 to version 1.9.19-201912131005

### LINKED ISSUE #369: getid3 PHP 7.4 warnings
The current bundled version of getid3 is not fully compatible with PHP 7.4 . [Latests versions 1.9.18 and 1.9.19](https://github.com/JamesHeinrich/getID3/releases) contain the necessary fixes.

Warning output:
```
PHP message: PHP Deprecated:  Array and string offset access syntax with curly braces is deprecated in ../s3-uploads/lib/getid3/getid3.php on line 554
PHP message: PHP Deprecated:  Array and string offset access syntax with curly braces is deprecated in ../s3-uploads/lib/getid3/getid3.php on line 555
```

luishdz1010:
Fixes 7.4 PHP incompatibilities

Fixes #369

### DIFF
--- a/lib/getid3/extension.cache.dbm.php
+++ b/lib/getid3/extension.cache.dbm.php
@@ -2,10 +2,9 @@
 
 /////////////////////////////////////////////////////////////////
 /// getID3() by James Heinrich <info@getid3.org>               //
-//  available at http://getid3.sourceforge.net                 //
+//  available at https://github.com/JamesHeinrich/getID3       //
 //            or https://www.getid3.org                        //
-//          also https://github.com/JamesHeinrich/getID3       //
-/////////////////////////////////////////////////////////////////
+//            or http://getid3.sourceforge.net                 //
 //                                                             //
 // extension.cache.dbm.php - part of getID3()                  //
 // Please see readme.txt for more information                  //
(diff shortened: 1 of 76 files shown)


AdrienPoupa:
Why not use composer for this?

luishdz1010:
> Why not use composer for this?

See https://github.com/humanmade/S3-Uploads/blob/master/s3-uploads.php#L47-L54 for the reasoning.

It seems that Core finally updated the getID3 library to a more recent version (1.19.18). Perhaps we don't need to include our own copy anymore?

AdrienPoupa:
I'm not questioning the reasoning behind not using WordPress' but they could include the lib with composer instead of versioning it

joehoyle:
good points to both @luishdz1010 and @AdrienPoupa.

- [ ] Confirm if we still need this newer version, or did WordPress update?
- [ ] Pull this in via composer if it's still needed.

austinpray:
@joehoyle vendoring/bringing this in with composer is not needed anymore. 1.9.18 (included in core) fixes the original reason why vendoring was required.

1.9.19 fixes some linting issues: but that would not be a concern if this was brought in with composer anyway. 

I would tag a breaking change that
- requires a recent version of WordPress
- removes vendored getid3

tillkruss:
@joehoyle: Here is the core changeset: https://core.trac.wordpress.org/changeset/46112

joehoyle:
Closed in favor of #391, thanks for the investigation and work @luishdz1010 @tillkruss  @austinpray @AdrienPoupa

Issue #325, opened 2019-05-25
PHP allow_url_fopen flag [off] makes plugin fail silently


vinerz:
We just did a major platform upgrade recently and S3 Uploads stopped working:

> Unable to create directory. Is its parent directory writable by the server?

I was almost giving up tonight when I had no more hair to rip off my head, but in a desperate move I noticed the `STREAM_IS_URL` flag at `stream_wrapper_register`:

https://github.com/humanmade/S3-Uploads/blob/f17559cba87e0a0a0191a747f62c74dbde60208d/inc/class-s3-uploads-stream-wrapper.php#L115

This, of course, clashes with our `php.ini` directive which disallows such requests.
The thing is: PHP fails completely silent when this is turned off, so I couldn't find the root cause fast enough.

I just enabled `allow_url_fopen` and it is now back working as it should.

It would be great to insert this as a known caveat at the README for future developers.

Thanks for this amazing plugin!

Preen:
Thanks for this!!

VlasicVojislav:
Going to get fixed via:
https://github.com/humanmade/S3-Uploads/pull/705

PR #383, opened 2020-02-27
Add url_fopen option checking to requirements

dmgawel:
The plugin uses `stream_wrapper_register` with `STREAM_IS_URL` flag, thus requiring `allow_url_fopen` to be enabled. It is a common security practice among hosting providers to disable this option as a precautionary measure. In such a setup the user only sees following error:

> Unable to create directory. Is its parent directory writable by the server?

This PR adds a check during plugin init and provides appropriate error message when `allow_url_fopen` is disabled.

Related issue: #325

### DIFF
--- a/s3-uploads.php
+++ b/s3-uploads.php
@@ -75,6 +75,14 @@ function s3_uploads_check_requirements() {
 
 		return false;
 	}
+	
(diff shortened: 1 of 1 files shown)


joehoyle:
Great, thanks @dmgawel

PR #393, opened 2020-03-29
Documenting how to manually install dependencies

andrewspear:


joehoyle:
In https://github.com/humanmade/S3-Uploads/pull/415 we will now be automatically including a built file with deps included in each release, so closing this PR out as it's no longer needed.

PR #398, opened 2020-04-17
Add support for Private Attachments

joehoyle:
This is an experiment to support private / unpublished uploads. These files are pushed to S3 with a private ACL, and then URLs are generated and signed when pages are generated, with a time-based expiry.

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,6 +146,14 @@ Private uploads can be transitioned to public by calling `S3_Uploads::set_attach
 S3_Uploads::get_instance()->set_attachment_files_acl( 15, 'public-read' );
 ```
 
+The default expiry for all private file URLs is 24 hours. You can modify this by using the `s3_uploads_private_attachment_url_expiry` WordPress filter. The value can be any string interpreted by `strtotime`. For example:
+
+```php
+add_filter( 's3_uploads_private_attachment_url_expiry', function ( $expiry ) {
+	return '+1 hour';
+} );
+```
+
 ## Cache Control
 
 You can define the default HTTP `Cache-Control` header for uploaded media using the
--- a/README.md
+++ b/README.md
@@ -134,7 +134,7 @@ Note: as either `<from>` or `<to>` can be S3 or local locations, you must specif
 
 WordPress (and therefor S3 Uploads) default behaviour is that all uploaded media files are publicly accessible. In certain cases which may not be desireable. S3 Uploads supports setting S3 Objects to a `private` ACL and providing temporarily signed URLs for all files that are marked as private.
 
-S3 Uploads does not make assumptions for provide UI for marking attachments as private, instead you should integrate the `s3_uploads_is_attachment_private` WordPress filter to control the behaviour. For example, to mark _all_ attachments as private:
+S3 Uploads does not make assumptions or provide UI for marking attachments as private, instead you should integrate the `s3_uploads_is_attachment_private` WordPress filter to control the behaviour. For example, to mark _all_ attachments as private:
 
 ```php
 add_filter( 's3_uploads_is_attachment_private', '__return_true' );
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ It's focused on providing a highly robust S3 interface with no "bells and whistl
 
 ## Requirements
 
-- PHP >= 5.5
+- PHP >= 7.1
 - WordPress >= 5.3
 
 ## Getting Set Up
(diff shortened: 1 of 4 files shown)


joehoyle:
Left to do:

- Support for detecting the object-acl on upload. Right now this is set in the sreamwrapper, so there is no Attachment ID in context, need to somehow pull that out.
- Right now there's no UI or way to mark an attachment as private.
- The expiry time of the signed URLs should be configurable.

inc/class-s3-uploads.php
```diff
@@ -277,4 +348,142 @@ public function copy_image_from_s3( $file ) {
...
 	}
+
+	/**
+	 * Check if the attachment is private.
+	 *
+	 * @param integer $attachment_id
+	 * @return boolean
+	 */
+	public function is_private_attachment( int $attachment_id ) : bool {
+		return apply_filters( 's3_uploads_is_attachment_private', true, $attachment_id );
```
stuartshields, line 359: @joehoyle I think this should be:
```
return apply_filters( 's3_uploads_is_attachment_private', false, $attachment_id );
```

Instead of forcing private attachments on everyone by default, we should allow them to opt in via:
```
 add_filter( 's3_uploads_is_attachment_private', '__return_true' );
```

Unless the intended use case you have is for it to be on by default. If that's the case should we update the above filter to in the readme:
```
 add_filter( 's3_uploads_is_attachment_private', '__return_false' );
```

joehoyle:
@stuartshields hey, wanted to check how you got on with this -- I noticed you add a couple of commits. Is this branch working for you now?

stuartshields:
@joehoyle The branch is working for me, I pushed it to our dev server to make sure it works there. The only issue we're facing is Tachyon but I think updating the Tachyon version should resolve this.

README.md
```diff
@@ -130,6 +130,30 @@ wp s3-uploads cp <from> <to>
...
+add_filter( 's3_uploads_is_attachment_private', '__return_true' );
+```
+
+Private uploads can be transitioned to public by calling `S3_Uploads::set_attachment_files_acl( $id, 'public-read' )` or vica-versa. For example:
+
+```php
+S3_Uploads::get_instance()->set_attachment_files_acl( 15, 'public-read' );
+```
+
+The default expiry for all private file URLs is 24 hours. You can modify this by using the `s3_uploads_private_attachment_url_expiry` WordPress filter. The value can be any string interpreted by `strtotime`. For example:
```
stuartshields, line 149: @joehoyle This needs to be updated to reflect the changes you made. 24 hours needs to be 6 hours.

joehoyle: Ok updated!

inc/class-s3-uploads.php
```diff
@@ -277,4 +352,163 @@ public function copy_image_from_s3( $file ) {
...
+		$locations = array_map( [ $this, 'get_s3_location_for_path' ], $files );
+		$s3 = $this->s3();
+		$commands = [];
+		foreach ( $locations as $location ) {
+			$commands[] = $s3->getCommand( 'putObjectAcl', [
+				'Bucket' => $location['bucket'],
+				'Key' => $location['key'],
+				'ACL' => $acl,
+			] );
+			try {
```
Nikschavan, line 390: Should the try/catch block be outside the foreach? Right now it seems to be executing the `CommandPool` on each command individually as soon as the command is added to the commands array.

Nikschavan, line 390: Should the try/catch block be outside of foreach? Right now it seems to be executing the CommandPool on each command individually as soon as the command is added to the commands array.

Nikschavan: Moved the comment in right line number - https://github.com/humanmade/S3-Uploads/pull/398/files#r422976931

joehoyle: Whoops! nice catch :D

inc/class-s3-uploads.php
```diff
@@ -162,6 +174,54 @@ public function get_original_upload_dir() {
...
+	}
+
+	/**
+	 * Reverse a file path in the uploads directory to the params needed for S3.
+	 *
+	 * @param string $url
+	 * @return array
+	 */
+	public function get_s3_location_for_path( string $path ) : ?array {
+		$parsed = parse_url( $path );
```
Nikschavan, line 215: Use `wp_parse_url` instead. This is also flagged by hm-linter, So it will fix that as well.

```suggestion
		$parsed = wp_parse_url( $path );
```

joehoyle: I fixed this in #401 , so think I'll just leave the commit there.

joehoyle:
@stuartshields did you do an implementation for updating / replacing the URLs in post content? Thinking that should probably be added to S3 Uploads in this PR, as things are kinda broken without it.

This is an open source repo, but maybe you can DM me with any specific links to code where you have implemented that.

inc/class-s3-uploads.php
```diff
@@ -277,4 +352,168 @@ public function copy_image_from_s3( $file ) {
...
+	}
+
+	/**
+	 * Add the S3 signed params to an image src array.
+	 *
+	 * @param array|false $image
+	 * @param integer $post_id
+	 * @return array|false
+	 */
+	public function add_s3_signed_params_to_attachment_image_src( $image, int $post_id ) {
```
stuartshields, line 479: @joehoyle So we didn't take into the account that `wp_get_attachment_image_src` can accept both `array|false` as seen here https://developer.wordpress.org/reference/functions/wp_get_attachment_image_src/ I actually came across this on the project.

joehoyle: Good catch, thanks @stuartshields

Issue #143, opened 2017-01-03
PDF thumbnails introduced in WP 4.7 do not work


ikappas:
When a user uploads a PDF on WP 4.7 it creates a PDF thumbnail attachment.

This does not work with S3 Uploads due to a suffix `[0]` being applied to the filename in order to fetch the first page of the pdf in WP_Image_Editor_Imagick::pdf_setup as seen below:
```php
/**
	 * Sets up Imagick for PDF processing.
	 * Increases rendering DPI and only loads first page.
	 *
	 * @since 4.7.0
	 * @access protected
	 *
	 * @return string|WP_Error File to load or WP_Error on failure.
	 */
	protected function pdf_setup() {
		try {
			// By default, PDFs are rendered in a very low resolution.
			// We want the thumbnail to be readable, so increase the rendering DPI.
			$this->image->setResolution( 128, 128 );

			// Only load the first page.
			return $this->file . '[0]';
		}
		catch ( Exception $e ) {
			return new WP_Error( 'pdf_setup_failed', $e->getMessage(), $this->file );
		}
	}
```

S3 Client cannot handle this suffix and throws an exception. A quick workaround is to add the following function in S3_Uploads_Image_Editor_Imagick :
```php
	/**
	 * {@inheritDoc}
	 */
	protected function pdf_setup() {
		$filename = parent::pdf_setup();
		if ( is_wp_error( $filename ) ) {
			return $filename;
		}
		return substr( $filename, 0, strlen( $filename ) - 3 );
	}
```

ikappas:
@rmccue Any news on this issue?

rmccue:
ping @joehoyle

ikappas:
@joehoyle @rmccue Any progress on this?

joehoyle:
I made #236 with a fix, let me know if this works for you @ikappas

PR #400, opened 2020-04-29
213: Fix Travis CI builds and use wp-phpunit/wp-phpunit to run phpunit tests

### LINKED ISSUE #213: Update Travis CI config
CI is failing due to configuration problems.

Nikschavan:
Fixes #213


joehoyle:
Hey @Nikschavan thanks for this! As it happens I've also put some work into the tests, to get a single docker image to run it all instead, so we can more easily run it locally and the like. That isn't quite ready, so I tihnk there would be benefit to merging this if it's already working, but I see the build is still failing here.

Nikschavan:
This PR is still WIP, I will change this to draft until then.

Currently, only one of the [tests seems to be failing](https://travis-ci.com/github/humanmade/S3-Uploads/builds/162635859#L353) as to previously Travis was not starting test due to different configuration errors. 

The failing test is `Test_S3_Uploads::test_generate_attachment_metadata_for_pdf` which seems to be added in to https://github.com/humanmade/S3-Uploads/issues/143

I am still debugging this test right now. It appears the function `wp_get_image_editor()` is returning `WP_Error` when testing for the thumbnail generation for the PDF files. If you have any ideas on this do let me know.

joehoyle:
Hmm that might be because imamagick is not set up to handle PDF images. I had to do work in my branch to fix it. I'll push up my unit test changes to another PR so we can compare approaches.

joehoyle:
@Nikschavan I've pushed my approach up in https://github.com/humanmade/S3-Uploads/pull/402, which uses a self-contained docker image as the test environment. That will also install imagick and have the PDF extensions

Nikschavan:
The approach in #402 sounds nice as it also does not require uploading to an actual s3 bucket during tests.

Do you think it's worth the effort to continue working on #400? I am happy to close this PR in favor of #402.

joehoyle:
@Nikschavan ok cool. Yes, I thikn that will prove to be the better aproach. Are you ok to tinker with docker too? I think it would be great to switch to wp-phpunit away from SVN in the docker image buidling. Also, if you wanted to have a bash at setting up travis to run the image that would be great. I've been using this command to run tests:

`docker run --rm -v $PWD/../../:/code -it s3-uploads-unit-tests`.

We can build the image and upload it to dockerhub too, so maybe the image will be more like `humanmade/s3-uploads-unit-tests`.

I think we _could_ also split the dockerfile into 2 images: one that is a wordpress-unit-testing ready to go image, then inherit from that file/image in an s3-uploads specific image which just adds minio etc to it.

Nikschavan:
Sure, I will work on #402 to move that one forward and close #400.

PR #402, opened 2020-04-29
Run tests via docker image

joehoyle:
Rather than needing to install PHP, all modules, imagemagick, and S3 APis locally, docker would be much metter suited to running out tests. This way we can run an S3 replacement (minio) instead of requiring a live integration with S3. This is better for performance and isolation of tests. It also means we have a clearly defined, consistent envirinment for tests for PHP configuration.

Still to do:

- [x] Add travis config for running tests via docker
- [x] switch to wp-phpunit https://github.com/wp-phpunit/wp-phpunit


rmccue:
Dumb Q: rather than extending the container, could you just have a standalone minio container here? Seems like you could then use the [off-the-shelf minio container](https://docs.min.io/docs/minio-docker-quickstart-guide)

joehoyle:
We could yes, though that would require setting up docker-compose and also coming up with a declarative way to create the s3 bucket (which can be done via a somewhat odd docker-composer service declaration like we do in local-server). Advantage would be that we don’t need to build a custom image of course

--

Joe Hoyle

CTO
Human Made

rmccue:
> that would require setting up docker-compose

Does it? Can't we just add another line to the Travis file?

joehoyle:
Well then we have to set up the shared networks too, which compose handles. To be able to run these on other hosts too, like macOs then it starts to get more complex to do manually like that

--

Joe Hoyle

CTO
Human Made

rmccue:
From my investigation and further discussion, you don't need docker-compose, you can communicate via the IP easily enough:

```
docker run -d --rm --name=minio minio/minio server /data
MINIO_IP=$(docker inspect -f '{{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}' minio)
docker run -e "MINIO_IP=$MINIO_IP" humanmade/plugin-tester
```

Then Minio is available on `$MINIO_IP:9000`

joehoyle:
Ok cool, we’ll need to create the bucket too, I’m not sure if the minio image comes with mc, or if that’s another container too

rmccue:
Based on https://github.com/minio/minio/issues/4769 seems like if we mount the FS in, that'll work?

```
mkdir -p /tmp/minio/bucket
docker run ... -v /tmp/minio:/data minio/minio server
```

PR #403, opened 2020-04-29
Use wp-phpunit + docker to run tests in ci

Nikschavan:
- Locally commands can be run with `composer test` (After the main image is pulled once)
- I have kept composer dependencies outside of the image thinking it can be easier to switch the WordPress version easily. Should these dependencies be inside the image?
- Right now the docker image is pushed to `nikschavan/s3-uploads-unit-tests` after the image contents are finalized it can be pushed from the humanmade account.

WIP -
- Using separate images for base wordpress-phpunit configuration and minio. Can we use docker-compose for this to combine service with these two images?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 .idea/
 *.phar
 vendor/
+wordpress/
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,29 +3,24 @@
 # Tell Travis CI we're using PHP
 language: php
 
+services:
+  - docker
+
+cache:
+  timeout: 1000
+  directories:
+    - vendor
+
 notifications:
   email: false
 
 php:
   - 7.0
 
-env:
-  - WP_VERSION=latest WP_MULTISITE=0
-
-matrix:
-  include:
-    - php: 7.0
-      env: WP_VERSION=latest WP_MULTISITE=1
-
 # Clones WordPress and configures our testing environment.
 before_script:
-  - bash bin/install-wp-tests.sh wordpress_test root '' 127.0.0.1 $WP_VERSION
-  - printf "\n" | pecl install imagick
+  - docker pull nikschavan/s3-uploads-unit-tests
   - composer install
 
 script:
-  - phpunit --coverage-clover clover.xml
-
-after_script:
-  # Push coverage off to Codecov
-  - bash <(curl -s https://codecov.io/bash)
+  - composer test
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ notifications:
   email: false
 
 php:
-  - 7.0
+  - 7.2
 
 # Clones WordPress and configures our testing environment.
 before_script:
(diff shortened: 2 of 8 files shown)


joehoyle:
Rather than docker-compose, I was thinking we should just extend the base image with “FROM: humanmade/Wordpress-tests” or whatever. So we are building a new docker image just for s3uploads

Nikschavan:
What about having composer dependencies built into the image or keeping them outside to make it easier to switch the WP Versions like it is right now?

joehoyle:
I was thinking maybe we could set up something to batch publish images for all version of Wordpress. Iirc the wp-php init version needs to (roughly) match the Wordpress version. We _could_ pull them outside of the image, but it will mean slower docker run commands i think (unless we have some pointed volume to cache the data in). Rather than that I was thinking we just push a tag for each major version of Wordpress. The general idea of this approach was to have zero other deps, very “hard coded” everything in the image approach

joehoyle:
To that end, I think we would atleast want to make Wordpress version a build argument in the docker file

joehoyle:
@Nikschavan Ok, I pulled out the base part of the image into github.com/humanmade/plugin-tester/, so then in https://github.com/humanmade/S3-Uploads/pull/402/files#diff-9235a7bdb1ca9268681c0e56f3f3609bR1 I switched to using it, so it's a bit simpler now.

> I have kept composer dependencies outside of the image thinking it can be easier to switch the WordPress version easily. Should these dependencies be inside the image?

I think these should be in the image, so it's simpler to use, I was thinking we could do a tag-per-image like `plugin-tester:5.4` etc then.

Nikschavan:
This is nice, I was thinking we will need to move the docker images outside of this repo.

to maintain multiple images, I am thinking we can use something like this - https://github.com/WordPress/wpdev-docker-images/blob/master/update.php
Maybe use these images as a base or just use a similar approach to build the matrix of WordPress/PHP Versions.

Can you outline in an issue on `humanmade/plugin-tester` as a general idea of how many PHP/WP Versions should be maintained? This will help me (and others) pick this work later.

Nikschavan:
I did not reach to this when working on this last week, although I thought I might hit as a blocker when trying to use `wp-phpunit` inside the docker image - 

If we have unit tests related composer dependencies in `/vendor` inside the image and we also mount the plugin with its separate dependencies, will that cause any conflict/issue since the plugin and the image both have the same `/vendor` directory?

joehoyle:
Ok, I copied your travis / composer.json script to #402 , which is now passing in https://travis-ci.com/github/humanmade/S3-Uploads/builds/163698090!

At the moment, `phpunit` is specific as a dev dep, but `wp-phpunit` is not (that one is bundled in the container). I am not 100% sure the best way to approach it -- it would be fine having wp-phpunit as a dep too, though I think `phpunit` is not a dep of `wp-phpunit` (for some reason!) so you have to have both.

`wordpress` and `wp-phpunit` are WordPress version specific, so I feel like it makes more sense for them to be part of the docker image which is tagged with the WordPress version. PHPUnit is less so (though you do need specific versions of PHP Unit with specific versions of wp-phpunit`, so there is still that.

I think this conversation would be best continued on humanmade/plugin-tester though

Nikschavan:
Closing this as part of this is merged to #402

PR #404, opened 2020-05-03
Use https instead of http in the plugin URI field

tomjn:
A minor bit of polish adding an `s` to a protocol handler


joehoyle:
Thanks!

Issue #368, opened 2019-12-21
Installation error?


AdrienPoupa:
I tried to install the plugin by doing

```
composer require humanmade/s3-uploads
```

at the root of my WordPress installation. Then I enabled it with WP CLI (that I also installed with Composer):

```
vendor/wp-cli/wp-cli/bin/wp plugin activate s3-uploads
```

But I was getting the following error because the dependencies could not be loaded:

```
Warning: require_once(/var/www/wp-content/plugins/s3-uploads/vendor/autoload.php): failed to open stream: No such file or directory in /var/www/wp-content/plugins/s3-uploads/s3-uploads.php on line 21
```

I had to composer install within the plugin directory whereas I thought it would retrieve the dependencies from the root of my WP installation? or am I missing something?

Thanks

Edit: looks like a duplicate of https://github.com/humanmade/S3-Uploads/issues/313

codeagencybe:
Any update yet on this issue? 
I'm having exactly same problem. 

`[21-Jan-2020 11:19:14 UTC] PHP Warning:  require_once(/var/www/demo/wp-content/plugins/S3-Uploads-2.2.1/vendor/autoload.php): failed to open stream: No such file or directory in /var/www/demo/wp-content/plugins/S3-Uploads-2.2.1/s3-uploads.php on line 21
[21-Jan-2020 11:19:14 UTC] PHP Fatal error:  require_once(): Failed opening required '/var/www/demo/wp-content/plugins/S3-Uploads-2.2.1/vendor/autoload.php' (include_path='.:/usr/local/lib/php') in /var/www/demo/wp-content/plugins/S3-Uploads-2.2.1/s3-uploads.php on line 21
`

x0rzkov:
I could not install the plugin neither

rmccue:
Hi. If you install S3 Uploads via Composer, you need to ensure you're loading Composer's autoload file yourself, and before S3 Uploads is loaded.

This error only happens if you are loading via Composer **and** are not correctly loading Composer's autoload file.

AdrienPoupa:
No, you are the one not doing it correctly... As a user, I expect to be able to install the plugin via Composer from the root of my WP installation, enable it with WP CLI and it should work out of the box.

Having to cd in the plugin's directory to manually install the dependencies is a non sense, and the number of users reacting to this ticket proves it.

rmccue:
[...]

PR #413, opened 2020-05-21
Update installation instructions

rmccue:
See discussion on #368.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,10 @@ It's focused on providing a highly robust S3 interface with no "bells and whistl
 composer require humanmade/s3-uploads
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -40,11 +40,13 @@ It's focused on providing a highly robust S3 interface with no "bells and whistl
 composer require humanmade/s3-uploads
 ```
 
+**Note:** [Composer's autoloader](https://getcomposer.org/doc/01-basic-usage.md#autoloading) must be loaded before S3 Uploads is loaded. We recommend loading it in your `wp-config.php` before `wp-settings.php` is loaded.
```
roborourke, line 43:
```suggestion
**Note:** [Composer's autoloader](https://getcomposer.org/doc/01-basic-usage.md#autoloading) must be loaded before S3 Uploads is loaded. We recommend loading it in your `wp-config.php` before `wp-settings.php` is loaded as shown below.

```php
require_once __DIR__ . '/vendor/autoload.php';
```
```

PR #417, opened 2020-06-10
Different directory names between site and S3 bucket

erasmusa:
### My use case

I already have an S3 bucket containing all the Wordpress Multisite's media files but the uploads directory is not called "uploads". The website is also behind an existing CDN so the URL path needs to remain the same which includes that "/uploads/" section of the path but the S3 bucket should not have that. I'm also not using the URL rewrites and configured nginx to reverse proxy to the S3 private bucket by signing the headers in the request.. The only missing bit was that the plugin needed to make this distinction in folder/directory names between the URL and the S3 bucket.

This is my attempt at achieving this goal.

To make this work add the following line to `wp-config.php`

`define('S3_UPLOADS_DROP_UPLOADS_DIR', true);`

Feel free to improve on this - I had to get something up and running quickly.

### DIFF
--- a/inc/class-s3-uploads.php
+++ b/inc/class-s3-uploads.php
@@ -452,7 +452,7 @@ class S3_Uploads {
 	 * @param integer $post_id
 	 * @return string
 	 */
-	public function add_s3_signed_params_to_attachment_url( string $url, int $post_id ) : string {
+	public function add_s3_signed_params_to_attachment_url( string $url, $post_id ) : string {
 		if ( ! $this->is_private_attachment( $post_id ) ) {
 			return $url;
 		}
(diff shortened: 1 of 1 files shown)


noplanman:
@erasmusa Do you mind sharing the header signing part of the reverse proxy? I've been trying to achieve this too but am running into this issue: https://github.com/anomalizer/ngx_aws_auth/issues/17

Do you have a different method or is there an existing config or how-to/tutorial you can point me to?
Thanks!

erasmusa:
@noplanman Sorry for this late reply. Things have been a bit hectic on my end of the woods. I'll write up some instructions for you on how to achieve the header signing in nginx - I just need to retrace my steps first ;-)

noplanman:
@erasmusa Just a gentle ping, as I still haven't managed to get it to work without issues :confused:

joehoyle:
Closing this as stale, but feel free to update if this is still valid.

PR #422, opened 2020-07-03
Fix slow call to scandir() when uploading files

### LINKED ISSUE #373: Extremely slow _wp_handle_upload()
Hey guys,

Love this plugin - been using it for years.

I've stumbled across a problem with `_wp_handle_upload()` where it's taking way too long to handle the uploaded file. I discovered this when trying to import attachments programmatically with `media_handle_sideload()`.

I've wrote a small script on a fresh WP install to test this out with an 12KB image.

```php
$starttime_upload = microtime(true);
_wp_handle_upload($file, ['test_form' => false, 'test_size' => false], current_time('mysql'), 'wp_handle_sideload');
$endtime_upload = microtime(true) - $starttime_upload;

echo "took {$endtime_upload}s to handle upload";
```

Which outputs:
```
took 2.6300148963928s to handle upload
```

If I disable S3-Uploads plugin I get:
```
took 0.029120206832886s to handle upload
```

Just to check I ran `wp s3-uploads verify` and it verified in around 2-300ms.

Not sure if this is expected or not - would appreciate any input you guys might have on this.

joehoyle:
Please for give me for what I have done.

Fixes #373.

### DIFF
--- a/inc/class-stream-wrapper.php
+++ b/inc/class-stream-wrapper.php
@@ -589,6 +589,32 @@ class Stream_Wrapper {
 
 		$this->openedBucketPrefix = $params['Key'];
 
+		// WordPress attempts to scan whole directories via wp_unique_filename(), which can be very slow
+		// when there are thousands of files in a single uploads sub directory. This is due to behaviour
+		// introduced in https://core.trac.wordpress.org/changeset/46822/. Essentially when a file is uploaded,
+		// it's not enough to make sure no filename already exists (and append a `-1` to the end), because
+		// image sizes of that image could also conflict with already existing files too. Because image sizes
+		// (in the form of -800x600.jpg) can be arbitrary integers, it's not possible to iterate the filesystem
+		// for all possible matching / colliding file names. WordPress core uses a preg-match on all files that
+		// might conflict with the given filename.
+		//
+		// Fortunately, we can make use of S3 arbitrary prefixes to optimize this query. The WordPress regex
+		// done via _wp_check_existing_file_names() is essentially `^$filename-...`, so we can modify the prefix
+		// to include the filename, therefore only return a subset of files from S3 that are more likely to match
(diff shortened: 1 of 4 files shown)


joehoyle:
@rmccue could I get a 👍 / 👎  today on this, as this is currently blocking some production errors.

joehoyle:
@rmccue I added support for the filter too here.

inc/class-stream-wrapper.php
```diff
@@ -583,11 +583,45 @@ public function dir_opendir( $path, $options ) {
...
+		// to include the filename, therefore only return a subset of files from S3 that are more likely to match
+		// the preg_match() call.
+		//
+		// Essentially, wp_unique_filename( my-file.jpg ) doing a `scandir( s3://bucket/2019/04/ )` will actually result in an s3
+		// listObject query for `s3://bucket/2019/04/my-file` which means even if there are millions of files in `2019/04/` we only
+		// return a much smaller subset.
+		//
+		// Anyone reading this far, brace yourselves for a mighty horrible hack.
+		$backtrace = debug_backtrace( 0, 3 ); // phpcs:ignore PHPCompatibility.FunctionUse.ArgumentFunctionsReportCurrentValue.NeedsInspection
+		if ( isset( $backtrace[1]['function'] ) && $backtrace[1]['function'] === 'scandir' && isset( $backtrace[2]['function'] ) && $backtrace[2]['function'] === 'wp_unique_filename' ) {
```
rmccue, line 620: Does this need to bail if it hit the filter callback above, or is that already handled?

joehoyle: This won't be triggered, as the backtrace doesn't reflect our own `scandir` call.

rmccue: Gotcha.

PR #433, opened 2020-07-22
Allow UPLOADS constant to be defined

jimwebb:
Allow upload file path to be modified by the UPLOADS constant, sometimes set by sites.


joehoyle:
Hmm I'm not sure we want to support `ms_files_rewriting`. I'm going to close this out, but thanks for the PR.

PR #451, opened 2020-10-12
Expose filter to allow injecting S3 StreamWrapper cache instance

lukaspawlik:
Expose WordPress filter to allow other plugins to easily inject `\Aws\S3\StreamWrapper` cache instance.

This filter would make easy integration of solutions like https://github.com/lukaspawlik/s3-uploads-cache without doing hacks of re-registering StreamWrapper in your plugin.


joehoyle:
Thanks @lukaspawlik, this looks valuable. Since #475, the interface has slightly changed here, so this needs refreshing. Could you handle? Thanks!

Issue #416, opened 2020-06-09
Should we remove type hinting on wp_get_attachment_image_src because of WordPress?


stuartshields:
@joehoyle So I'm working on the Secure Media stuff and came across this error:
```
Uncaught Error: Argument 2 passed to S3_Uploads::add_s3_signed_params_to_attachment_image_src() must be of the type integer, string given
```

This is happening because we expect $post_id to only be `int`: 
https://github.com/humanmade/S3-Uploads/blob/master/inc/class-s3-uploads.php#L486

However in WordPress if you do something like:
```
wp_get_attachment_image_src( get_post_thumbnail_id( get_the_ID() ), 'full', false )
```

`get_post_thumbnail_id` returns int|string:
```
(int|string) Post thumbnail ID or empty string if the post does not exist.
```
reference: https://developer.wordpress.org/reference/functions/get_post_thumbnail_id/

Thoughts on this? As it may break other projects code.

joehoyle:
Hrm is this string(0) @stuartshields? Per https://3v4l.org/qotFg, it looks like this error should only be triggered if it's passed `""`, because `"1"` will be cast to an int.

joehoyle:
So, yes, looks like we will need to remove this typehint, unfortunately,.

Koli14:
> So, yes, looks like we will need to remove this typehint, unfortunately,.

I have created a PR for this: https://github.com/humanmade/S3-Uploads/pull/458

PR #401, opened 2020-04-29
Switch to using PHP namespaces

joehoyle:
To keep inline with the HM coding standards and consistency with other Human Made code, I'd like to swich S3 Uploads to namespaces, and fix up any outstanding PHP formatting issues with the HM PHPCS.

PR #458, opened 2020-10-26
#416 - remove typehint in add_s3_signed_params_to_attachment_image_src

Koli14:
This PR fix this issue: https://github.com/humanmade/S3-Uploads/issues/416

### DIFF
--- a/inc/class-s3-uploads.php
+++ b/inc/class-s3-uploads.php
@@ -483,7 +483,7 @@ class S3_Uploads {
 	 * @param integer $post_id
 	 * @return array|false
 	 */
-	public function add_s3_signed_params_to_attachment_image_src( $image, int $post_id ) {
+	public function add_s3_signed_params_to_attachment_image_src( $image, $post_id ) {
 		if ( ! $image ) {
 			return $image;
 		}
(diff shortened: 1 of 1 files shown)


joehoyle:
I believe we fixed this in https://github.com/humanmade/S3-Uploads/pull/401, which contains a bunch of refactoring. Just waiting on a review from @rmccue to push that forward.

Koli14:
> I believe we fixed this in #401, which contains a bunch of refactoring. Just waiting on a review from @rmccue to push that forward.

I don't see, if it's changed: https://github.com/humanmade/S3-Uploads/blob/dce05fbaf1515a849e905afa604620d22623acb6/inc/class-plugin.php#L583
Although it's still might be fixed...

joehoyle:
Ahh ok good point, I thought that code had been refactored too, but perhaps not.

inc/class-s3-uploads.php
```diff
@@ -483,7 +483,7 @@ public function add_s3_signed_params_to_attachment_url( string $url, int $post_i
 	 * @param integer $post_id
```
joehoyle, line 483: Could the phpdoc be updated too to reflect the possible types?

joehoyle:
Hey @kodie I closed this in favour of https://github.com/humanmade/S3-Uploads/pull/474, as there was a big amount of code churn from #401

PR #464, opened 2020-10-29
PUB-4335: expose attachment metadata to S3 upload filters

laingawbl:


laingawbl:
whoops! nobody mind me, I did that prematurely...

PR #474, opened 2020-12-08
Loosen type signature add_s3_signed_params_to_attachment_image_src

### LINKED ISSUE #416: Should we remove type hinting on wp_get_attachment_image_src because of WordPress?

joehoyle:
Fixes #416

### DIFF
--- a/inc/class-plugin.php
+++ b/inc/class-plugin.php
@@ -581,11 +581,11 @@ class Plugin {
 	 * Add the S3 signed params to an image src array.
 	 *
 	 * @param array{0: string, 1: int, 2: int}|false $image
-	 * @param integer $post_id
+	 * @param integer|"" $post_id The post id, due to WordPress hook, this can be "", so can't just hint as int.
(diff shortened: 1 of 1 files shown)


stuartshields (approved):
@joehoyle Change looks good to me. Approved.

PR #475, opened 2020-12-12
Lazy load s3 client

joehoyle:
All requests right now will init the s3client, which causes a lot of file i/o in loading the aws sdk, and a fair bit of CPU to crunch through the sdk init.

Now we have this much better typed, I think we can do what we were never able to do in https://github.com/humanmade/S3-Uploads/pull/128


rmccue:
@joehoyle can you update this from master with latest lint changes?

joehoyle:
@rmccue updated!

rmccue (approved):
Nice, easy!

PR #479, opened 2020-12-17
Publish type coverage to Psalm Shepherd

joehoyle:
This only published from the master branch so there's not much we can do other than merge and hope!


rmccue (approved):
yolo!

PR #491, opened 2021-02-15
Cache presigned URLs fetched from AWS SDK

goldenapples:
The AWS SDK by default uses a blocking synchronous request to fetch presigned request URLs for resources. This means adding substantial overhead in cases like the WordPress media library where up to 40 requests have to be made in series before returning a page of results from the admin-ajax query.

There are a couple separate functional improvements I'd like to make here, which I'll be making as separate pull requests. This one seems like the simplest improvement, and doesn't change the basic functionality here: if we get a presigned URL that's valid for a given period of time, we should be able to cache the URL for that same period of time so as not to require a new API request to sign that URL every time the attachment is previewed.

### DIFF
--- a/inc/class-plugin.php
+++ b/inc/class-plugin.php
@@ -557,24 +557,28 @@ class Plugin {
 		if ( ! $path ) {
 			return $url;
 		}
-		$cmd = $this->s3()->getCommand(
-			'GetObject',
-			[
+
+		$presigned_url = wp_cache_get( $url, 's3_uploads_presigned_url' );
+
+		if ( ! $presigned_url ) {
+			$cmd = $this->s3()->getCommand('GetObject', [
 				'Bucket' => $path['bucket'],
 				'Key' => $path['key'],
-			]
-		);
+			]);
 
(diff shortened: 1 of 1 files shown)


joehoyle:
Hmm when I implemented this, getting a pre-signed URL did not require a network request, I'm wondering if that's changed -- as I understand this process, it should not be needed.

joehoyle:
Looking at https://github.com/aws/aws-sdk-php/blob/1ab4f565744aea8e9560ee8aaffb19b87238c7d5/src/S3/S3Client.php#L458, I wondering if perhaps it's just the fetching of credentials that is the remote request?

I don't see any other remote requests in the implementation

goldenapples:
Alright let me take a deeper look under the hood and see what's going on. Maybe the network request is just to get the signing provider, in which case we should be able to fetch that once and reuse. There are a few memoized functions in the AWS SDK which look like they do that.

goldenapples:
Closing this PR, as you're correct - the request to get a presigned URL isn't a remote request. (I really don't understand how expiration and invalidation works for S3 assets, in this case, but I can figure that out separately.) Caching the presigned URL like I'm doing here doesn't solve anything.

PR #493, opened 2021-02-18
Save filesize with attachment metadata on upload

### LINKED ISSUE #492: Idea: store filesize attachment meta value on upload.
There are a few places in core, like `wp_prepare_attachment_for_js`, that do a filesystem stat check for attachments to get the filesize in bytes:

https://github.com/WordPress/WordPress/blob/master/wp-includes/media.php#L3969-L3975

This can be short-circuited if the "filesize" property is set in the attachment meta array.

Setting this property seems as though it would eliminate the need for several AWS S3 requests which happen as part of the `query-attachments` ajax action.  Is it feasible to set it on uploading attachments?

goldenapples:
Sets the "filesize" property in the attachment metadata when initially uploading images, so that it doesn't have to be queried for in the ajax response on `wp_prepare_attachment_for_js()`.

Fixes #492.

### DIFF
--- a/inc/class-plugin.php
+++ b/inc/class-plugin.php
@@ -110,6 +110,7 @@ class Plugin {
 		add_filter( 'wp_resource_hints', [ $this, 'wp_filter_resource_hints' ], 10, 2 );
 
 		add_action( 'wp_handle_sideload_prefilter', [ $this, 'filter_sideload_move_temp_file_to_s3' ] );
+		add_filter( 'wp_generate_attachment_metadata', [ $this, 'set_filesize_in_attachment_meta' ] );
 
 		add_action( 'wp_get_attachment_url', [ $this, 'add_s3_signed_params_to_attachment_url' ], 10, 2 );
 		add_action( 'wp_get_attachment_image_src', [ $this, 'add_s3_signed_params_to_attachment_image_src' ], 10, 2 );
@@ -129,6 +130,7 @@ class Plugin {
 		remove_filter( 'upload_dir', [ $this, 'filter_upload_dir' ] );
 		remove_filter( 'wp_image_editors', [ $this, 'filter_editors' ], 9 );
 		remove_filter( 'wp_handle_sideload_prefilter', [ $this, 'filter_sideload_move_temp_file_to_s3' ] );
+		remove_filter( 'wp_generate_attachment_metadata', [ $this, 'set_filesize_in_attachment_meta' ] );
 
 		remove_action( 'wp_get_attachment_url', [ $this, 'add_s3_signed_params_to_attachment_url' ] );
 		remove_action( 'wp_get_attachment_image_src', [ $this, 'add_s3_signed_params_to_attachment_image_src' ] );
@@ -376,7 +378,6 @@ class Plugin {
 
 		return $editors;
 	}
-
 	/**
 	 * Copy the file from /tmp to an s3 dir so handle_sideload doesn't fail due to
 	 * trying to do a rename() on the file cross streams. This is somewhat of a hack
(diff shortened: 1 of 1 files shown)


joehoyle:
Nice, I had assumed core already did this -- do you know why it doesn't?

inc/class-plugin.php
```diff
@@ -396,6 +397,33 @@ public function filter_sideload_move_temp_file_to_s3( array $file ) {
...
+	 * longer to complete against s3 uploads than if they were performed with a
+	 * local filesystem.i
+	 *
+	 * Saving the filesize in the attachment metadata when the image is
+	 * uploaded allows core to skip this stat when retrieving and formatting it.
+	 *
+	 * @param array $metadata Attachment metadata.
+	 * @return array Attachment metadata array, with "filesize" value added.
+	 */
+	function set_filesize_in_attachment_meta( $metadata ) {
```
joehoyle, line 415:
```suggestion
	function set_filesize_in_attachment_meta( array $metadata ) : array {
```

joehoyle:
The unit tests are failing here, because `metadata['file']`, surprisingly is not always set. It seems that's only for images, videos for example never get that set (https://core.trac.wordpress.org/browser/tags/5.6.1/src/wp-admin/includes/media.php#L3532). It's possible to get the file instead with `get_attached_file()` rather than reading from the metadata for that value. TIL

inc/class-plugin.php
joehoyle: This project is using Psalm too, so once the unit tests are fixed, I'd imagine we'll need to add the array shape decelerating like elsewhere in the file.

goldenapples:
> I had assumed core already did this -- do you know why it doesn't?

I don't think there was a reason it wasn't added. The logic to check for "filesize" in attachment meta was added in 4.4 ([ticket](https://core.trac.wordpress.org/ticket/33214)) to support external image libraries and I guess the assumption was that those plugins had to implement attachment metadata themselves. Looks like there's a ticket open to set filesize in this metadata array in core - https://core.trac.wordpress.org/ticket/49412 - and if that gets merged this workaround can be removed.

inc/class-plugin.php
```diff
@@ -396,6 +397,32 @@ public function filter_sideload_move_temp_file_to_s3( array $file ) {
...
+	 * uploaded allows core to skip this stat when retrieving and formatting it.
+	 *
+	 * @param array{file?: string} $metadata      Attachment metadata.
+	 * @param int                  $attachment_id Attachment ID.
+	 * @return array{file?: string, filesize: int} Attachment metadata array, with "filesize" value added.
+	 */
+	function set_filesize_in_attachment_meta( array $metadata, int $attachment_id ) : array {
+		$file = get_attached_file( $attachment_id );
+
+		if ( file_exists( $file ) ) {
```
joehoyle, line 419: I think this should return early if filesize is already set too, that will mean it's future compatible when images add it.

inc/class-plugin.php
```diff
@@ -396,6 +397,32 @@ public function filter_sideload_move_temp_file_to_s3( array $file ) {
...
+	 * result, operations like `wp_prepare_attachments_for_js' take substantially
+	 * longer to complete against s3 uploads than if they were performed with a
+	 * local filesystem.i
+	 *
+	 * Saving the filesize in the attachment metadata when the image is
+	 * uploaded allows core to skip this stat when retrieving and formatting it.
+	 *
+	 * @param array{file?: string} $metadata      Attachment metadata.
+	 * @param int                  $attachment_id Attachment ID.
+	 * @return array{file?: string, filesize: int} Attachment metadata array, with "filesize" value added.
```
joehoyle, line 414: This should also be `filesize?` as it's possible the array won't contain it (if the file doesn't exist). I'd have thought Psalm would have caught that 🤔

joehoyle (changes requested):

goldenapples:
@joehoyle Could you help me understand why the build isn't passing here?

The error is 
```
Running Psalm...

Problem parsing /code/psalm.xml:
  Error on line 2:
    Element '{https://getpsalm.org/schema/config}psalm', attribute 'requireVoidReturnType': The attribute 'requireVoidReturnType' is not allowed.
``` 
but I'm not familiar with what that means, and I think I'm following the style from the documentation and elsewhere in the file.

inc/class-plugin.php
goldenapples: Yeah, I was thinking that too. Adding it now.

inc/class-plugin.php
goldenapples: Makes sense, but that didn't seem to fix the build either :thinking:

joehoyle:
Ahh doh it seemed I inadvertently switched to a different version of Psalm ( ` : void` on functions that don't return anything is now required). Will fix up.

inc/class-plugin.php
goldenapples: Fixed in 4c50bf1

inc/class-plugin.php
```diff
@@ -396,6 +397,34 @@ public function filter_sideload_move_temp_file_to_s3( array $file ) {
...
+	 * Saving the filesize in the attachment metadata when the image is
+	 * uploaded allows core to skip this stat when retrieving and formatting it.
+	 *
+	 * @param array{file?: string} $metadata      Attachment metadata.
+	 * @param int                  $attachment_id Attachment ID.
+	 * @return array{file?: string, filesize?: int} Attachment metadata array, with "filesize" value added.
+	 */
+	function set_filesize_in_attachment_meta( array $metadata, int $attachment_id ) : array {
+		$file = get_attached_file( $attachment_id );
+		if ( ! $file ) {
```
joehoyle, line 418: @goldenapples `get_attached_file` _can_ return false, so I added the check here -- Psalm was complaining about that. Yey for type safety!

joehoyle (approved):

PR #494, opened 2021-02-19
Switch to fixed version of Psalm Plugin WordPress

joehoyle:
The latest version of Psalm-plugin-wordpress is Psalm 4, which needs some new upgrade steps. We'll do that, but first let's just switch back to the older version.


goldenapples (approved):
Looks good, thanks!

PR #497, opened 2021-03-11
Backport #493 to v2

goldenapples:
This PR manually backports the patch from #493 to the v2 branch.

See #492.


joehoyle (approved):
@goldenapples I might be wrong, but I _think_ we might not have used return type declarations as it's not supported by the version of PHP supported in 2x. Before merging could you just verify if we are using them elsewhere?

goldenapples:
@joehoyle you were right, there aren't any return types elsewhere in the plugin on v2-branch. I've removedf the one in this PR in 
ecd933e.

joehoyle:
awesome thanks

joehoyle:
tagged version 2.2.3 https://github.com/humanmade/S3-Uploads/releases/tag/2.2.3

PR #507, opened 2021-04-20
Don't install dev deps in manual install

joehoyle:
I added `workflow_dispatch` too, to be able to debug with https://github.blog/2021-04-15-work-with-github-actions-in-your-terminal-with-github-cli/


joehoyle:
Per https://github.com/humanmade/S3-Uploads/runs/2389523488?check_suite_focus=true this looks to be fixed up finally! We have to install the pcov extension too.

roborourke (approved):
`workflow_dispatch` looks pretty cool!

PR #512, opened 2021-04-30
[BC break] Keep Guzzle v6 in v2

karser:
This plugin is bundled with `aws-sdk-php` which requires `guzzlehttp/guzzle:^5.3.3|^6.2.1|^7.0`.
In version 2.2.1 there was `guzzlehttp/guzzle:6.5.5`.
But in 2.3.0  it's now `guzzlehttp/guzzle:7.2.0`.

I have other plugins that require guzzle v6 and they crashed after the update to v2.3.0.
Since in PHP it's not possible to have 2 classes loaded of a different version, because if the class is already loaded it will not load it twice.

Please downgrade the guzzle version, at least in v2.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,6 +20,7 @@
 	"type": "wordpress-plugin",
 	"require": {
 		"composer/installers": "~1.0",
-		"aws/aws-sdk-php": "~3.18"
-	 }
+		"aws/aws-sdk-php": "~3.18",
+		"guzzlehttp/guzzle": "^6.2"
+	}
 }


roborourke:
Are your other plugin installed via composer too? Or do they bundle Guzzle v6?

karser:
The same way as S3-Uploads: it bundles Guzzle v6 installed with composer

karser:
Maybe it should be a separate build with guzzle-v6? Like manual-install-guzzle-v6.zip?

roborourke:
If they’re installed via composer it should be able figure out a compatible set of packages. What happens if you add guzzle v6 as a requirement in your main composer file?

karser:
I don't have composer.json on the project level. The plugin zip files are downloaded and extracted, and that's it. So every plugin has its own `vendor` folder and autoloading.
Do you have global composer.json? How does it work for wordpress?

roborourke:
Ah I see. Yeah that makes things difficult. You have a couple of options but I'd recommend reading through this primer https://composer.rarst.net/

At the simplest level you can use a `composer.json` and add https://wpackagist.org/ as a repository, then you can install WP plugins / themes. Composer will automatically place packages of the type `wordpress-plugin` in the plugins directory for example.

joehoyle:
Close as old

PR #516, opened 2021-05-10
Update .gitignore

raj4rachit:


roborourke:
We don't want to commit the vendor directory, it's available as part of the built releases in the manual-install.zip file under the assets section here for example https://github.com/humanmade/S3-Uploads/releases/tag/2.3.0

The manual zip install release workflow is currently broken for 3.0 so we'll try and get that sorted as soon as we can.

Issue #515, opened 2021-05-07
Existing AWS SDK class causes fatal error because vendor/autoload.php is not required


modelm:
Version c6c23f1f98016e45ca275f288d5ec27513cabfec In inc/namespace.php `init()`:

```
	// Ensure the AWS SDK can be loaded.
	if ( ! class_exists( '\\Aws\\S3\\S3Client' ) ) {
		// Require AWS Autoloader file.
		require_once dirname( __DIR__ ) . '/vendor/autoload.php';
	}
```

On this install, this class does already exist because another plugin is using the AWS SDK. However the S3-Uploads plugin still needs to load /vendor/autoload.php to avoid this error:

```
Fatal error: Uncaught Error: Class 'S3_Uploads\Plugin' not found in /app/wordpress/wp-content/plugins/S3-Uploads/inc/namespace.php:32 Stack trace: #0 /app/wordpress/wp-includes/class-wp-hook.php(287): S3_Uploads\init('') #1 /app/wordpress/wp-includes/class-wp-hook.php(311): WP_Hook->apply_filters(NULL, Array) #2 /app/wordpress/wp-includes/plugin.php(484): WP_Hook->do_action(Array) #3 /app/wordpress/wp-settings.php(420): do_action('plugins_loaded') #4 /app/wordpress/wp-config.php(105): require_once('/app/wordpress/...') #5 /app/wordpress/wp-load.php(37): require_once('/app/wordpress/...') #6 /app/wordpress/wp-admin/admin.php(34): require_once('/app/wordpress/...') #7 /app/wordpress/wp-admin/plugins.php(10): require_once('/app/wordpress/...') #8 {main} thrown in /app/wordpress/wp-content/plugins/S3-Uploads/inc/namespace.php on line 32 
```

Just removing the condition seems to work fine. Why not always require the autoload file?

modelm:
To avoid a related error with WP CLI, the require line should be closer to the top of the file under `namespace` instead of inside `init()`.

PR #519, opened 2021-05-18
fix: always require autoload.php

modelm:
This fixes https://github.com/humanmade/S3-Uploads/issues/515, a fatal missing class error on the site as well as in WP CLI commands with the plugin active and the AWS SDK loaded from another source.

### DIFF
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -2,13 +2,9 @@
 
 namespace S3_Uploads;
 
-function init() {
-	// Ensure the AWS SDK can be loaded.
-	if ( ! class_exists( '\\Aws\\S3\\S3Client' ) ) {
-		// Require AWS Autoloader file.
-		require_once dirname( __DIR__ ) . '/vendor/autoload.php';
-	}
+require_once dirname( __DIR__ ) . '/vendor/autoload.php';
 
+function init() {
 	if ( ! check_requirements() ) {
 		return;
 	}


joehoyle:
We dont want to always include autoloader.php, as that should only be used in the case where you are installing s3-uploads completely standalone outside of your project using composer (with all deps installed to `/vender`). If you have another plugin or dependency that includes the aws-sdk, I'd recommend installed s3-uploads and that plugin via composer, so they can both share the same AWS SDK, and you'll get all the dependencies of both plugins.

modelm:
I understand, but maybe you should add a disclaimer to the readme that the plugin will throw a fatal error if you just install it from source and run composer install, and also have the SDK already loaded. It would be better to avoid the error IMO - I don't expect this type of error in this scenario as a plugin installer.

korridor:
@modelm We had the same problem and with your pull request WPCLI works again.

@joehoyle If I understand the problem correctly the following sentence in the readme is not correct:

> You can also git clone this repository, and run composer install in the plugin folder to pull in its dependencies.

joehoyle:
@korridor hmm I'm not sure-- I had though this issue was created because `composer install` was not run? The description states "AWS SDK loaded from another source." which I took to mean "outside of composer's autoloader".

modelm:
@korridor is correct. Steps to reproduce #515 on a site that already loads the SDK via another plugin:

1. git clone the plugin
2. composer install
3. activate plugin

In this scenario, this plugin's autoload.php file is never loaded because the condition only checks that the SDK class exists, which it does but not because this plugin loaded it. But this plugin always throws a fatal error as a result, until/unless its own autoload.php file is loaded, like I've done in this PR.

korridor:
@modelm This sounds like a very similar problem, but in my case I have no other plugin with the S3 SDK. I just created a issue with a more detailed description of my problem here: #530

gabriserra:
Works as a charm. I cannot see any issue with it. Maintainers, please consider this PR.

nickpelton:
Bump. Had this issue with the manual install build release as well. This PR fixes the issue. Please consider this PR.

MattWilliamsDev:
Alternatively, this fix could be applied by adding the following to `/s3-uploads.php` above the statement that requires `namespace.php`:

```php
// Ensure the AWS SDK can be loaded.
if ( ! class_exists( '\\Aws\\S3\\S3Client' ) ) {
	// Require AWS Autoloader file.
	require_once dirname( __FILE__ ) . '/vendor/autoload.php';
}

require_once __DIR__ . '/inc/namespace.php';

...
```

MattWilliamsDev:
Maybe @kovshenin can review this? I've seen his name pop up recently on commits and nobody else seems to look at this

rmccue:
@MattWilliamsDev Please do not ping people repeatedly across multiple issues, or we may need to limit your ability to comment on Human Made projects.

MattWilliamsDev:
@rmccue Hey it worked though! This is the first time it seems anyone has looked at this in quite some time. 

There doesn't seem to be a way around using composer with this package. There also doesn't seem to be a working installation for people that don't already have the AWS SDK running (like this PR aims to fix). These are problems for a lot of people and they're problems that also seem to lack documentation.

This solution only solves cases that still use composer, but that doesn't help for manual install because the dependencies aren't packaged with the tarball. Even when I install the AWS SDK manually and then add the steps to load the SDK without composer, this still fails because it still needs something to be autoloaded.

dscottboggs:
I can confirm that adding @MattWilliamsDev's patch to an installation done with `manual-install.zip` works. I'm not experienced with PHP so perhaps I'm missing something, but it looks like that solution only requires `vendor/autoload.php` if it needs to be, so is that not an acceptable solution?

roryheaney:
I can also confirm that @MattWilliamsDev patch works for the latest manual install.

rmccue:
Thanks for the PR! We're going to close this in favour of #644.

Issue #513, opened 2021-05-05
Must be an instance of S3_Uploads\AWS\Command, instance of Aws\Command given


techypaul:
When running:

`wp s3-uploads upload-directory /var/www/wordpress/wp-content/uploads/ uploads --verbose`

I recieve:

`PHP Fatal error:  Uncaught TypeError: Argument 1 passed to S3_Uploads\WP_CLI_Command::S3_Uploads\{closure}() must be an instance of S3_Uploads\AWS\Command, instance of Aws\Command given, called in /var/www/wordpress/vendor/aws/aws-sdk-php/src/S3/Transfer.php on line 395 and defined in /var/www/wordpress/wp-content/plugins/s3-uploads/inc/class-wp-cli-command.php:209`

If I modify the line 209:

`'before'      => function ( AWS\Command $command ) {`
to
`'before'      => function ( Aws\Command $command ) {`

It works fine. I have used this tool before without issues. I am thinking this is a change in SDK possibly, but please advise?

techypaul:
I realised v3 is released. I get this in composer show:

```
aws/aws-sdk-php                  3.180.0
humanmade/s3-uploads             3.0.0
```

If I set to v2 in composer, i.e.
`"humanmade/s3-uploads": ">=2.0 <3.0"`

I get:
```
aws/aws-sdk-php                  3.180.2
humanmade/s3-uploads             2.3.0
```

And everything is working. I will stick with v2 but given that it is the only change I am making, I leave this issue open as I don't understand the problem.

iamlili:
Hi,

Came here to report this exact bug

I also just installed the latest:
```
aws/aws-sdk-php                                   3.183.10 
humanmade/s3-uploads                              3.0.0  
```

HOWEVER - this is what worked for me

changed
`'before' => function ( AWS\Command $command ) {`
to
`'before' => function ( \AWS\Command $command ) {`

PR #521, opened 2021-06-01
Fix upload-directory command

iamlili:
- was using incorrect namespace
- this fixes https://github.com/humanmade/S3-Uploads/issues/513

### DIFF
--- a/inc/class-wp-cli-command.php
+++ b/inc/class-wp-cli-command.php
@@ -2,6 +2,7 @@
 
 namespace S3_Uploads;
 
+use Aws\Command;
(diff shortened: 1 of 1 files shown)


joehoyle:
Hmm wondering how this didn't get caught by the static analysis via psalm.

joehoyle:
Ah IIRC this is to do with the WP CLI stubs not being available so the WP CLI command does not get type checking.

joehoyle (approved):

joehoyle:
@iamlili thanks!

iamlili:
@joehoyle no probs!

PR #531, opened 2021-06-23
Remove extraneous files

dingo-d:
This PR will remove unnecessary files for production release (psalm, configuration files, etc.)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,5 +1,29 @@
+#
+# Exclude these files from release archives.
+# This will also make them unavailable when using Composer with `--prefer-dist`.
+# https://blog.madewithlove.be/post/gitattributes/
+#
 /tests export-ignore
-/bin export-ignore
-/.gitignore export-ignore
-/.travis.yml export-ignore
-/phpunit.xml.dist export-ignore
+/psalm export-ignore
+/.github export-ignore
+.gitattributes export-ignore
+.gitignore export-ignore
+.phpcsignore export-ignore
+.travis.yml export-ignore
+codecov.yml export-ignore
+phpunit.xml.dist export-ignore
+psalm.xml export-ignore
+verify.txt export-ignore
+
+#
+# Auto detect text files and perform LF normalization
+# http://davidlaing.com/2012/09/19/customise-your-gitattributes-to-become-a-git-ninja/
+#
+* text=auto
+
+#
+# The above will handle all files NOT found below
+#
+*.md text
+*.php text
+*.inc text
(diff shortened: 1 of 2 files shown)


.github/workflows/build-release.yml
```diff
@@ -26,7 +26,7 @@ jobs:
       uses: thedoctor0/zip-release@master
       with:
         filename: 'manual-install.zip'
-        exclusions: '*.git* *tests/ *tests/data/ *tests/* *tests/**/*'
+        exclusions: '*.git* *tests/ *tests/data/ *tests/* *tests/**/* *psalm/* *.phpcsignore *.yml *.xml *.xml.dist'
```
joehoyle, line 29: I'm not sure about not including XMl and yml files, though could be required by packages such as the aws-sdk etc

.gitattributes
```diff
@@ -1,5 +1,29 @@
...
+verify.txt export-ignore
+
+#
+# Auto detect text files and perform LF normalization
+# http://davidlaing.com/2012/09/19/customise-your-gitattributes-to-become-a-git-ninja/
+#
+* text=auto
+
+#
+# The above will handle all files NOT found below
```
joehoyle, line 25: Why did we need to add this section?

joehoyle (changes requested):

.github/workflows/build-release.yml
dingo-d: Usually, these are not bundled in the release. I know that in WPCS we add these to `.gitattributes`. These are all development files that are usually not pulled when Composer is used but are present when you clone the repo.

.gitattributes
dingo-d: Just a minor quality improvement if the package is pulled via Composer.

.github/workflows/build-release.yml
joehoyle: Right but that should be dependent on what the deps need. E.g. if we excluded `.json` files that _would_ bread a lot of things, so I don't know for sure that excluding .xml and .yml is a good idea. How do we know that no Yaml / XML hires are needed for all the deps to work correctly?

dingo-d: Do you mean, like from vendor files? I'm not sure this exclusion list would look at that (might be wrong tho).

But `.yaml`, `.xml`, `.neon` are usually all config files. I don't think any of the dependencies are using/bundling those.

I can remove the general file extensions, and add the specific files to exclude (config files), shouldn't be a problem.

PR #541, opened 2021-07-30
Load S3 Uploads at 0 priority, rather than 10

### LINKED ISSUE #540: Change priority for S3 Uploads loader
Currently, S3's init loads on `plugins_loaded` at priority 10. This means uploads aren't correctly set up prior to this, i.e. if alphabetically-earlier plugins run callbacks on `plugins_loaded`.

We should run this at a much earlier priority, such as `0`, since all it does is add further hooks.

Acceptance criteria:

* [ ] Calling `wp_upload_dir()` from a plugin on the `plugins_loaded` (default priority) hook should return the S3 URL

rmccue:
This allows other plugins on the default priority to load correctly.

Fixes #540.

### DIFF
--- a/s3-uploads.php
+++ b/s3-uploads.php
@@ -10,7 +10,7 @@ Author URI: https://hmn.md
 
 require_once __DIR__ . '/inc/namespace.php';
 
-add_action( 'plugins_loaded', 'S3_Uploads\\init' );
+add_action( 'plugins_loaded', 'S3_Uploads\\init', 0 );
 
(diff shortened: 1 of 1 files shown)


roborourke (approved):
Should note that this is a breaking change as other code could be removing or replacing this hook on the current priority

rmccue:
> Should note that this is a breaking change as other code could be removing or replacing this hook on the current priority

Indeed, but we just broke BC with the latest releases by switching to namespaces, so unlikely it'll break any real use cases.

PR #546, opened 2021-08-19
Remove misleading documentation from README.md

jezemery:
Documentation in README.md suggests you can create an IAM user using the `create-iam-user ` subcommand, however this subcommand does not actually exist.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,15 +98,7 @@ like so:
 wp s3-uploads verify
 ```
 
-You'll want to create a new IAM user for the S3-Uploads plugin, so you are not using admin level access keys on your site. S3-Uploads can create the IAM user for you and asign the correct permissions.
-
-```
-wp s3-uploads create-iam-user --admin-key=<key> --admin-secret=<secret>
-```
-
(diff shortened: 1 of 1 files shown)


rmccue:
Huh, this was removed in  9a188e4784903941cac358491a18a7864030656c. @joehoyle was that intentional?

jezemery:
> Huh, this was removed in [9a188e4](https://github.com/humanmade/S3-Uploads/commit/9a188e4784903941cac358491a18a7864030656c). @joehoyle was that intentional?

I would say, it's a very useful command to have. So if there's work required to reinstate it, I'm happy to take a look.

joehoyle:
This was removed intentionally, mostly because it's something we never used, so it became out of date and untested. @jezemery if you wanted to PR and test it, I think we could bring it back.

jezemery:
I'll take a look and see what I can do

joehoyle:
In the meantime, we can merge this

joehoyle (approved):

PR #566, opened 2022-02-22
Image Editor Imagick: Always return the WP_Error object of parent call

ocean90:
If `WP_Image_Editor_Imagick::_save()` returns a `WP_Error` but no temp filename exists the `WP_Error` should still be returned. This prevents a fatal error `Cannot use object of type WP_Error as array` for the `$copy_result = copy( $save['path'], $filename );` line.


joehoyle (approved):

joehoyle:
Thanks @ocean90 !

PR #570, opened 2022-03-14
Set a returned value for `get_original_upload_dir`

mitogh:
When `original_upload_dir` is empty make sure the default value
is returned as otherwise no value would be returned.

### DIFF
--- a/inc/class-plugin.php
+++ b/inc/class-plugin.php
@@ -253,7 +253,7 @@ class Plugin {
 	public function get_original_upload_dir() : array {
 
 		if ( empty( $this->original_upload_dir ) ) {
-			wp_upload_dir();
+			return wp_upload_dir();
 		}
 
 		/**


inc/class-plugin.php
```diff
@@ -253,7 +253,7 @@ public function get_s3_bucket_region() : ?string {
 	public function get_original_upload_dir() : array {
 
 		if ( empty( $this->original_upload_dir ) ) {
-			wp_upload_dir();
+			return wp_upload_dir();
```
joehoyle, line 256: @mitogh this was intentional. Calling `wp_upload_dir` should always set `$this->original_upload_dir`. Otherwise doing this will return the `s3://` dir potentially.

mitogh: Oh makes sense, I was under the assumption this was a default value not a setup helper, thanks for the context :)

PR #596, opened 2022-08-12
Improve media library performance in some situations

### LINKED ISSUE #520: Hook slowing down WP load time
Hi guys the hook `add_filter( 'upload_dir', [ $this, 'filter_upload_dir' ] )` located on https://github.com/humanmade/S3-Uploads/blob/master/inc/class-plugin.php#L106 is slowing down the whole website changing the URL's on load time.
Is there a faster way to rewrite those URL?, maybe using `.htaccess`

### LINKED ISSUE #593: Massively increased load times
Hi!

Great plugin but I'm having an issue with performance. Read similar issues like mine, none of them gave me the needed answer. In most of them it's assumed that there's a conflict with another plugin or theme. Tried to debug it, disable all the plugins except essential ones, switched to default WP theme, sadly it didn't help. 

I've created two copies of the website with minimal functionality to show the problem. There's a mobile app which is getting data from the website so the main target is REST API endpoints. Both of these copies are in the same server and the setup is also identical except for the plugin in question. 

Several examples: 

http://s3-disabled.invsbl.lt/wp-json/app/v2/pois 4-5s load time 
https://s3.invsbl.lt/wp-json/app/v2/pois 22-26s load time

http://s3-disabled.invsbl.lt/wp-json/app/v2/routes 1-2s load time 
https://s3.invsbl.lt/wp-json/app/v2/routes 4-5s load time 

There are 3 plugins enabled in s3 version: https://nimb.ws/WBon7l 

ACF, S3 and our custom plugin which inits CPTs, creates endpoints and uses ACF functions to expose data. No optimisations, anything with file directories, just common ACF functions. 

This plugin is working great apart from performance problem so it would be awesome to find the problem but I'm unable to do that. Do you have any idea what might be wrong? 

Thanks a lot. 

Similar issues:

https://github.com/humanmade/S3-Uploads/issues/148
https://github.com/humanmade/S3-Uploads/issues/520
https://github.com/humanmade/S3-Uploads/issues/466
https://github.com/humanmade/S3-Uploads/issues/395
https://github.com/humanmade/S3-Uploads/issues/350 
https://github.com/humanmade/S3-Uploads/issues/431

claudiulodro:
Closes #520 
Closes #593 

In my testing on some large sites with hundreds of GBs of images in S3, the long loading times when displaying the media library that issues report are due to the `s3://` protocol in the path, (and likely other plugins inadvertently creating network requests to S3). If I've understood the logic correctly, the `s3://` protocol is only needed when working with the filesystem, and displaying the media library doesn't need to access the filesystem (unless I've missed some important use-case). 

Not having the `s3://` protocol available during the ajax request that displays the media library dramatically improves the performance of the media library without appearing to cause any side effects in my testing.

## To test:
1. Before applying this patch, check the timing of the `admin-ajax` request that fetches the media library data when adding an image block or on the Media screen in WP Admin. In this test, it's 17s for that one request:
Screen Shot 2022-08-12 at 12 33 56 PM

3. Apply this patch, refresh the page. Verify it takes much less time for the `admin-ajax` request that fetches the media library data and the images still display nicely:
Screen Shot 2022-08-12 at 12 34 22 PM

4. Test uploading an image. Images should still upload nicely to S3.

### DIFF
--- a/inc/class-plugin.php
+++ b/inc/class-plugin.php
@@ -188,7 +188,7 @@ class Plugin {
 				// The s3:// protocol is not needed when displaying the media library and slows things down in some cases.
 				if ( isset( $_POST['action'] ) && 'query-attachments' === $_POST['action'] ) {
 					$dirs['path']    = $this->original_upload_dir['path'];
-					$dirs['basedir'] = $this->original_upload_dir['basedir'];;
+					$dirs['basedir'] = $this->original_upload_dir['basedir'];
 				}
 			}
 		}
--- a/inc/class-plugin.php
+++ b/inc/class-plugin.php
@@ -184,6 +184,12 @@ class Plugin {
 			} else {
 				$dirs['url']     = str_replace( $s3_path, $this->get_s3_url(), $dirs['path'] );
 				$dirs['baseurl'] = str_replace( $s3_path, $this->get_s3_url(), $dirs['basedir'] );
+
+				// The s3:// protocol is not needed when displaying the media library and slows things down in some cases.
+				if ( isset( $_POST['action'] ) && 'query-attachments' === $_POST['action'] ) {
+					$dirs['path']    = $this->original_upload_dir['path'];
+					$dirs['basedir'] = $this->original_upload_dir['basedir'];;
+				}
 			}
 		}
 


claudiulodro:
Thinking on this further, I wonder if it might be even better to only have `s3://` available in the context of media upload. It's also not needed on e.g. the site frontend.

joehoyle:
Hi @claudiulodro thanks for the PR, however I don't think this is the right solution. If you're seeing slowing requests to the media library because of the s3:// protocol, it likley means you have something in the code that is trying to read the s3:// files as part of that request (which default WordPress will not do). For example, maybe there's something trying to read the image files to calculate sizes in a plugin, or similar.

joehoyle:
As this is old I'm going to close, but please to update us if it seems like the above explanation doesn't make sense.

PR #603, opened 2022-09-08
Added s3_uploads_s3_client filter to be able to change the S3Client

SeBsZ:
We added the ability to filter the S3Client returned by the s3() function. This allows us to switch the S3Client and provide it with different credentials, region and endpoint. We have a site where not all files are stored on the same S3 provider, and using the sample function below then gives us the ability to load another set of credentials. It would be great if you could merge this functionality in the main repo.

Sample function:

```
add_filter('s3_uploads_s3_client', function ($client)
{
    //Create new S3Client with other params
    $params = [ 'version' => 'latest' ];
    $params['credentials']['key'] = self::$s3_key;
    $params['credentials']['secret'] = self::$s3_secret;
    $params['signature'] = 'v4';
    $params['region'] = self::$s3_region;
    $params['endpoint'] = self::$s3_endpoint;
    $params['use_path_style_endpoint'] = true;
    $params['debug'] = false;

    $client = null;
    $sdk = new Aws\Sdk( $params );
    $client = $sdk->createS3();

    return $client;
});
```


joehoyle:
@SeBsZ are you not able to do this via `s3_uploads_s3_client_params`?

SeBsZ:
Sorry, I didn't see your message before @joehoyle . We can't do this with `s3_uploads_s3_client_params` as that only allows overriding the credentials, signature version, region and proxy settings. We need to be able to switch the endpoint/bucket.

joehoyle:
@SeBsZ those should also be part of `params`, we do this for example in Altis Local Server: https://github.com/humanmade/altis-local-server/blob/bc5916aa6c04f329933d12f415f7ea574166d68b/inc/namespace.php#L30-L35

PR #622, opened 2023-02-20
fix additional image sizes paths

neojp:
When trying to update the ACL on my images, I noticed the thumbnails weren't getting updated. While going through the rabbit hole I noticed the S3 batch commands were sending the following keys:

```
array (size=5)
  0 => string 's3://bucket-name/uploads/2023/02/unsplash_waxDxYM2XI4.jpg' (length=63)
  1 => string 's3://bucket-name/uploadsunsplash_waxDxYM2XI4-300x223.jpg' (length=62)
  2 => string 's3://bucket-name/uploadsunsplash_waxDxYM2XI4-1024x760.jpg' (length=63)
  3 => string 's3://bucket-name/uploadsunsplash_waxDxYM2XI4-150x150.jpg' (length=62)
  4 => string 's3://bucket-name/uploadsunsplash_waxDxYM2XI4-768x570.jpg' (length=62)
```

And they should be like this:

```
array (size=5)
  0 => string 's3://bucket-name/uploads/2023/02/unsplash_waxDxYM2XI4.jpg' (length=63)
  1 => string 's3://bucket-name/uploads/2023/02/unsplash_waxDxYM2XI4-300x223.jpg' (length=71)
  2 => string 's3://bucket-name/uploads/2023/02/unsplash_waxDxYM2XI4-1024x760.jpg' (length=72)
  3 => string 's3://bucket-name/uploads/2023/02/unsplash_waxDxYM2XI4-150x150.jpg' (length=71)
  4 => string 's3://bucket-name/uploads/2023/02/unsplash_waxDxYM2XI4-768x570.jpg' (length=71)
```

### DIFF
--- a/inc/class-plugin.php
+++ b/inc/class-plugin.php
@@ -551,7 +551,7 @@ class Plugin {
 		$meta = wp_get_attachment_metadata( $attachment_id );
 		if ( isset( $meta['sizes'] ) ) {
 			foreach ( $meta['sizes'] as $size => $sizeinfo ) {
-				$files[] = $uploadpath['basedir'] . $sizeinfo['file'];
+				$files[] = $uploadpath['path'] . '/' . $sizeinfo['file'];
 			}
 		}
 


inc/class-plugin.php
```diff
@@ -551,7 +551,7 @@ public static function get_attachment_files( int $attachment_id ) : array {
 		$meta = wp_get_attachment_metadata( $attachment_id );
 		if ( isset( $meta['sizes'] ) ) {
 			foreach ( $meta['sizes'] as $size => $sizeinfo ) {
-				$files[] = $uploadpath['basedir'] . $sizeinfo['file'];
+				$files[] = $uploadpath['path'] . '/' . $sizeinfo['file'];
```
joehoyle, line 554: Hmm `path` will return the current month directory. I guess I had thought that `$sizeinfo['file']` was a relative path to the basedir, but that doesn't seem to be the case. I think it might need to be something like `basedir( $main_file ) . '/' . $sizeinfo['file']`?

joehoyle (commented):
Thankss @neojp! I think we never ran into this as we use Tachyon for image downsizing from the original in our stack.

inc/class-plugin.php
neojp: You were right, I definitely noticed the dates not matching after a while. I had to change the code a bit, but haven't had a chance revisit this ticket yet.

joehoyle:
Closing as stale / no updates

PR #636, opened 2023-05-22
Replace "/" with "\" on windows machines

rasmuswinter:
When using the local stream wrapper on Windows machines, need to use "\\" directory separator instead of "/", otherwise `getLocalPath` always returns empty string


joehoyle (approved):

joehoyle:
Makes sense, thanks!

PR #641, opened 2023-06-04
docs: suggest uploading via the WP Admin console

blimmer:
This process was simpler for me than unzipping in the plugins directory and worked great!


rmccue:
Thanks for the PR! We're deprecating this installation method (#644), so will close out this PR.

PR #648, opened 2023-08-08
Fixes: Fatal Error when running PHP lint command

michaelbragg:
This PR fixes a linting error when running PHP's lint across the Imagick stub file.

When scanning our projects, including the S3 Uploads plugin, it was identified when running the lint command on new PHP versions.

System:
- PHP 7.4|8.0|8.1

Error:

```bash
Fatal error: Namespace declaration statement has to be the very first statement or after any declare call in the script in psalm/stubs/imagick.php on line 11
```

Recreate:
`php -l psalm/stubs/imagick.php`

Expect:
`No syntax errors detected in psalm/stubs/imagick.php`

Another option would be to exclude the `./psalm` directory from being included in the release via the `.gitattributes` file.

### DIFF
--- a/psalm/stubs/imagick.php
+++ b/psalm/stubs/imagick.php
@@ -1,11 +1,13 @@
 <?php
 
-/**
- * @method Imagick clone() (PECL imagick 2.0.0)<br/>Makes an exact copy of the Imagick object
- * @link https://php.net/manual/en/class.imagick.php
- */
-class Imagick implements Iterator, Countable {
+namespace {
+	/**
+	 * @method Imagick clone() (PECL imagick 2.0.0)<br/>Makes an exact copy of the Imagick object
+	 * @link https://php.net/manual/en/class.imagick.php
+	 */
+	class Imagick implements Iterator, Countable {
 
+	}
 }
 
 namespace WP_CLI {


michaelbragg:
Is there any update on a review of this pull request? Thanks.

tomjn:
I'd wrangle it myself but I'm unfamiliar with this project, and it doesn't look like @joehoyle has had the opportunity to look at it, nor has it affected paying clients.

Looking at the actual change, I'm curious why the solution is a `namespace { }` and not to move the WP CLI stub causing the problem out of the imagick stub and into a separate WP CLI file?

michaelbragg:
Thanks @tomjn.

> Looking at the actual change, I'm curious why the solution is a `namespace { }` and not to move the WP CLI stub causing the problem out of the imagick stub and into a separate WP CLI file?

That would be an equally valid fix. My initial idea was to make the least changes 😄. Happy to rework the PR. Does that help in getting it moving?

joehoyle:
Sorry for the delay, I think this makes sense!

joehoyle (approved):

michaelbragg:
Thanks @joehoyle, any chance of a cheeky suffix release for those of us who use composer?

PR #655, opened 2023-10-21
Use get_temp_dir instead of sys_get_temp_dir

ajakov:
I had a problem on one of the WordPress installation where directory returned by `sys_get_temp_dir` was not writeable by php and changing `WP_TEMP_DIR` did not help (because the plugin was not taking that into account).  
Changing `sys_get_temp_dir` to Wordpress function `get_temp_dir` helped.

### DIFF
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -146,7 +146,7 @@ function set_wp_privacy_exports_dir( string $dir ) {
 	if ( strpos( $dir, 's3://' ) !== 0 ) {
 		return $dir;
 	}
-	$dir = get_temp_dir() . '/wp_privacy_exports_dir/';
+	$dir = get_temp_dir() . 'wp_privacy_exports_dir/';
 	if ( ! is_dir( $dir ) ) {
 		mkdir( $dir );
 		file_put_contents( $dir . 'index.html', '' ); // @codingStandardsIgnoreLine FS write is ok.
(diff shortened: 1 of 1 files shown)


inc/namespace.php
```diff
@@ -136,7 +136,7 @@ function after_export_personal_data() {
  *
  * We don't want to use the default uploads folder location, as with S3 Uploads this is
  * going to the a s3:// custom URL handler, which is going to fail with the use of ZipArchive.
- * Instead we set to to sys_get_temp_dir and move the fail in the wp_privacy_personal_data_export_file_created
+ * Instead we set to to WPs get_temp_dir and move the fail in the wp_privacy_personal_data_export_file_created
```
tomjn, line 139: Would an apostrophe make sense here? Or remove "WPs" entirely?

```suggestion
 * Instead we set to to WP's get_temp_dir and move the fail in the wp_privacy_personal_data_export_file_created
```

ajakov: Yes, I forgot it, thank you!

inc/namespace.php
```diff
@@ -146,7 +146,7 @@ function set_wp_privacy_exports_dir( string $dir ) {
 	if ( strpos( $dir, 's3://' ) !== 0 ) {
 		return $dir;
 	}
-	$dir = sys_get_temp_dir() . '/wp_privacy_exports_dir/';
+	$dir = get_temp_dir() . '/wp_privacy_exports_dir/';
```
jerico, line 149:
```
wp> sys_get_temp_dir() 
=> string(4) "/tmp"
wp> get_temp_dir() 
=> string(5) "/tmp/"
```

`get_temp_dir` adds a trailing slash, though as far as I know PHP treats it as a single slash

ajakov: That's true. Thank you for pointing that out. I've changed that line now.

mikelittle (approved):
Looks good to me. 
Thank you, @ajakov

PR #661, opened 2024-01-11
fix readdir bug from PHP 8 substr change

sc0tth0lden:
As of PHP 8.0, substr will return empty string rather than false in certain circumstances.

From https://www.php.net/manual/en/function.substr.php : 
8.0.0 	The function returns an empty string where it previously returned false. 

Due to this change, this implementation of readdir in S3-Uploads behaves incorrectly. In particular, this common pattern:
while (($file = readdir($dh)) !== false){
which is used in gravityforms, becomes an infinite loop since the while loop just keeps getting an empty string rather than false.

This can lead to a crazy high AWS S3 list line item on your monthly bill. I recommend merging in this change right away to avoid this situation.

### DIFF
--- a/inc/class-stream-wrapper.php
+++ b/inc/class-stream-wrapper.php
@@ -724,9 +724,11 @@ class Stream_Wrapper {
 
 		// Remove the prefix from the result to emulate other stream wrappers.
 		$retVal = $this->openedBucketPrefix
-			  ? substr( $result, strlen( $this->openedBucketPrefix ) )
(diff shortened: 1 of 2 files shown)


tomjn (approved):
While I'm not super familiar with this plugin the explanation and the check make sense, and also catches when $result is '', I've approved but would like another engineer more familiar to check things before merge

mikelittle (approved):
Looks good. Thanks, Scott.

PR #663, opened 2024-01-25
Isolate dependencies using strauss and a vendor-prefix folder

felipeelia:


rmccue:
FYI, we don't intend on doing this in S3 Uploads itself as we intentionally use a shared AWS SDK (and S3 Uploads is not designed to be a traditional plugin). Feel free to maintain a fork though :)

felipeelia:
Hi @rmccue, yeah, I imagined that would be the case. The PR was meant to be open in my fork. Thanks for jumping in!

PR #682, opened 2025-01-29
Add note about checksums

### LINKED ISSUE #681: WordPress `6.7` compatibility?
I just upgraded to WordPress `6.7.1` and started seeing the following error when trying to upload media:

> _The uploaded file could not be moved to uploads/2025/01._

Looks like it's coming from here:

[wp-admin/includes/file.php:1033](https://github.com/WordPress/WordPress/blob/5c35a5f4c58fbb73f43e11bfed88677d40ad94f6/wp-admin/includes/file.php#L1033)

I debugged some of the runtime values around the error:

```php
Array (
    [action] => wp_handle_upload
    [file] => Array (
        [name] => IMG_1710.jpeg
        [full_path] => IMG_1710.jpeg
        [type] => image/jpeg
        [tmp_name] => /tmp/phpDLpiAN
        [error] => 0
        [size] => 1751699
    )
    [filename] => IMG_1710.jpeg
    [new_file] => s3://wp-prod/uploads/2025/01/IMG_1710.jpeg
    [type] => image/jpeg
    [uploads] => Array (
        [path] => s3://wp-prod/uploads/2025/01
        [url] => https://uploads.example.com/uploads/2025/01
        [subdir] => /2025/01
        [basedir] => s3://wp-prod/uploads
        [baseurl] => https://uploads.example.com/uploads
        [error] =>
    )
)
```

Notes:

- I am using [CloudFlare R2](https://developers.cloudflare.com/r2/) as my S3-compatible storage backend.
- I have another website using an older version of WordPress with the same setup which still works fine. 
- I tried disabling all other plugins / mu-plugins, but still see the same issue
- I tried downgrading WordPress back to the last version it was working on _(`6.6.2`)_, but still saw the issue..

Anybody else experiencing the same?

rmccue:
S3-compatible APIs may be broken with the latest AWS SDK, but can be worked around by disabling the checksums.

Fixes #681.

### DIFF
--- a/README.md
+++ b/README.md
@@ -229,6 +229,17 @@ add_filter( 's3_uploads_s3_client_params', function ( $params ) {
 } );
 ```
 
+**Note:** As of AWS SDK 3.337, S3 uses a [new type of integrity protection](https://github.com/aws/aws-sdk-php/issues/3062) which is not supported by all third-party S3-compatible APIs. If you experience errors, you may need to disable the new checksum functionality:
+
+```php
+add_filter( 's3_uploads_s3_client_params', function ( $params ) {
+	// ...
+	$params['request_checksum_calculation'] = 'when_required';
+	$params['response_checksum_validation'] = 'when_required';
+	return $params;
+} );
+```
+
 ## Temporary Session Tokens
 
 If your S3 access is configured to require a temporary session token in addition to the access key and secret, you should configure the credentials using the following code:


archon810:
Thank you for the tip on how to fix this issue.

Just ran into it with Cloudflare R2. Even though they closed https://www.cloudflarestatus.com/incidents/t5nrjmpxc1cj as fixed, we still ran into errors when using `exif_imagetype`.

```
PHP Warning:  Stream is not seekable in ...plugins/s3-uploads/inc/class-stream-wrapper.php on line 980
PHP Warning:  exif_imagetype(s3://...bla.png): Failed to open stream: &quot;S3_Uploads\Stream_Wrapper::stream_open&quot; call failed in .../plugins/...bla.php on line 71
```

`getimagesize` as well.

Whose fault is it at this point (as in, who do we bug to fix it properly)? S3-Uploads, aws-sdk-php (we're using 3.337.3), or Cloudflare?

rmccue:
> Whose fault is it at this point (as in, who do we bug to fix it properly)? S3-Uploads, aws-sdk-php (we're using 3.337.3), or Cloudflare?

S3 Uploads uses the AWS SDK, and the AWS SDK is only designed for AWS services. Cloudflare and others offer an "S3 compatible" API, so it's their responsibility to ensure they're compatible. In this case, I'm not sure there's really an issue anyway; it's just one of the items of config you'll need to set per-provider, just like changing the URL.

Note that S3 Uploads doesn't _officially_ support any non-AWS services, but we don't want to make it intentionally difficult to use them (indeed, we use Minio for local development environments already).

joehoyle (approved):

joehoyle:
Ta

Issue #634, opened 2023-05-19
Plugin crashes with improper input


Firmware-Repairman:
The function `get_files_for_unique_filename_file_list()` returns a boolean (`false`) if the parameter passed to `scandir()` is not a directory. This causes the filter where it is used to crash. If `scandir()` returns false the function should return an empty array.

*Function*
```
	/**
	 * Override the files used for wp_unique_filename() comparisons
	 *
	 * @param array|null $files
	 * @param string $dir
	 * @return array
	 */
	public function get_files_for_unique_filename_file_list( ?array $files, string $dir, string $filename ) : array {
		$name = pathinfo( $filename, PATHINFO_FILENAME );
		// The s3:// streamwrapper support listing by partial prefixes with wildcards.
		// For example, scandir( s3://bucket/2019/06/my-image* )
		return scandir( trailingslashit( $dir ) . $name . '*' );
	}
```

*Error trace*
> Fatal error: Uncaught Error: Return value of S3_Uploads\Plugin::get_files_for_unique_filename_file_list() must be of the type array, bool returned in /srv/htdocs/wp-content/plugins/s3-uploads/inc/class-plugin.php on line 668
>
>Call stack:
>
>S3_Uploads\Plugin::get_files_for_unique_filename_file_list()
>wp-includes/class-wp-hook.php:308
>WP_Hook::apply_filters()
>wp-includes/plugin.php:205
>apply_filters()
>wp-includes/functions.php:2643
>...

rasmuswinter:
Duplicate of issue #630

Fixed in PR #637

tomjn:
As a stopgap, this will unhook the function with this issue and replace it with the fixed version, just be sure to remove it once fixed:

```php
// Remove the broken version.
add_action( 'init', function () {
	remove_filter( 'pre_wp_unique_filename_file_list', [ \S3_Uploads\Plugin::get_instance(), 'get_files_for_unique_filename_file_list' ], 10, 3 );
} );

// Add a temporary fix.
add_filter( 'pre_wp_unique_filename_file_list', function ( ?array $files, string $dir, string $filename ) : array {
	$name = pathinfo( $filename, PATHINFO_FILENAME );
	// The s3:// streamwrapper support listing by partial prefixes with wildcards.
	// For example, scandir( s3://bucket/2019/06/my-image* )
	$scandir = scandir( trailingslashit( $dir ) . $name . '*' );
	if ( $scandir === false ) {
[...]

Issue #630, opened 2023-04-30
Uploads fail using S3_UPLOADS_USE_LOCAL


henriquevcosta:
I just added this plugin in a Bedrock multisite environment running on Lando and because I'm in a local dev environment I'm using S3_UPLOADS_USE_LOCAL=true but am having an error that I really can't understand.

I appreciate that my webroot isn't really standard (/app/src/bedrock/web) but everything else seems to be working. At first glance I thought it could be permissions, but actually the plugin is able to create the folder structure and in a shell inside the container I can create files with that user.

```

[Sun Apr 30 06:40:32.910833 2023] [php:warn] [pid 1155] [client 172.21.0.2:53062] PHP Warning:  opendir(/app/src/bedrock/web/app/uploads/sites/2/s3/fake-bucket/uploads/sites/2/2023/04/lanoxin*): Failed to open directory: No such file or directory in /app/src/bedrock/web/app/mu-plugins/s3-uploads/inc/class-local-stream-wrapper.php on line 515, referer: https://cenas.mywebsite.lndo.site/wp/wp-admin/post.php?post=3&action=edit
[Sun Apr 30 06:40:32.910932 2023] [php:warn] [pid 1155] [client 172.21.0.2:53062] PHP Warning:  scandir(s3://fake-bucket/uploads/sites/2/2023/04/lanoxin*): Failed to open directory: "S3_Uploads\\Local_Stream_Wrapper::dir_opendir" call failed in /app/src/bedrock/web/app/mu-plugins/s3-uploads/inc/class-plugin.php on line 668, referer: https://cenas.mywebsite.lndo.site/wp/wp-admin/post.php?post=3&action=edit
[Sun Apr 30 06:40:32.910951 2023] [php:warn] [pid 1155] [client 172.21.0.2:53062] PHP Warning:  scandir(): (errno 2): No such file or directory in /app/src/bedrock/web/app/mu-plugins/s3-uploads/inc/class-plugin.php on line 668, referer: https://cenas.mywebsite.lndo.site/wp/wp-admin/post.php?post=3&action=edit
[...]

PR #687, opened 2025-03-31
PR: Scandir fix for #634 and #630

ramonfincken:
Needs array as return as per code 

Fixes
https://github.com/humanmade/S3-Uploads/issues/634
and
https://github.com/humanmade/S3-Uploads/issues/630

### DIFF
--- a/inc/class-plugin.php
+++ b/inc/class-plugin.php
@@ -674,7 +674,7 @@ class Plugin {
 		// The s3:// streamwrapper support listing by partial prefixes with wildcards.
 		// For example, scandir( s3://bucket/2019/06/my-image* )
 		$scandir = scandir( trailingslashit( $dir ) . $name . '*' );
-		if( $scandir === false ) {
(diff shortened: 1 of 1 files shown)


tomjn (approved):

tomjn:
Noting that this fixes this issue for me, CI is failing due to warnings about the PHP version from the AWS SDK.

From the PHP docs for `scandir`:

> Returns an [array](https://www.php.net/manual/en/language.types.array.php) of filenames on success, or [false](https://www.php.net/manual/en/reserved.constants.php#constant.false) on failure. If directory is not a directory, then boolean [false](https://www.php.net/manual/en/reserved.constants.php#constant.false) is returned, and an error of level [E_WARNING](https://www.php.net/manual/en/errorfunc.constants.php#constant.e-warning) is generated.

If no files are found or that folder does not exist, then it returns `false` and triggers the return type failure.

joehoyle (approved):

PR #695, opened 2025-05-30
Add Stream support for audit logs of all ACL sets

Rayhatron:
Add [Stream](https://wordpress.org/plugins/stream/) support for audit logs of all ACL sets done by S3 Uploads to files of an attachment. This will help when any investigations need to be done in regards to why an attachment's files have a certain ACL.

We can add support for more audit logging WordPress plugins if there's a need. Starting with support for Stream as such a need has come up and instead of doing an integration solely in the projects that have that need, doing it in S3 Uploads allows others to benefit.

The audit logging feature is opt in and will only work if Stream plugin is activated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,7 +127,7 @@ Note: as either `<from>` or `<to>` can be S3 or local locations, you must specif
 
 ## Private Uploads
 
-WordPress (and therefor S3 Uploads) default behaviour is that all uploaded media files are publicly accessible. In certain cases which may not be desireable. S3 Uploads supports setting S3 Objects to a `private` ACL and providing temporarily signed URLs for all files that are marked as private.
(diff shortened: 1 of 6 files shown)


joehoyle:
@Rayhatron thanks, though I'm wondering if we might be able to make this an "addon" plugin rather than in the core of S3 Uploads. I'm not sure currently at least that we want this in the base plugin. Are the hooks available for this to be a standalone plugin?

Rayhatron:
@joehoyle this can be released now. The add-on plugin repo has been setup https://github.com/humanmade/s3-uploads-audit and the plugin published on packagist https://packagist.org/packages/humanmade/s3-uploads-audit for installing via composer.

PR #703, opened 2025-06-08
Fix implicit nullable parameter deprecations in PHP 8.4

philo23:
Just two small, one character fixes for [PHP 8.4's deprecated notice around implicit nulls](https://wiki.php.net/rfc/deprecate-implicitly-nullable-types).

```
PHP Deprecated:  S3_Uploads\Stream_Wrapper::register(): Implicitly marking parameter $cache as nullable is deprecated, the explicit nullable type must be used instead in /plugins/s3-uploads/inc/class-stream-wrapper.php on line 119
```

This fix should be compatible all the way back to PHP 7.1


joehoyle:
Thanks!

Issue #715, opened 2025-08-07
increment version number with release


BrookeDot:
There are quite a few issues currently opened relating to the version number not being bumped when a new version is released dating back to at least version 2. 
#525
#651 
#684 
#676 
#617 

I'll create a PR for a manual bump for 3.0.10, however, this issue is open to start a discussion around a programmatic way to bump the version number.

joehoyle:
Yup would be great, currently the flow us to use github releases to create the tag, so I think it would need to something like a github action to hook into "create_release" and I guess delete and re-create the tag after bumping... could be dodgy though.

PR #716, opened 2025-08-07
Bump version to 3.0.10

BrookeDot:
Bumps the version to 3.0.10. 

See: #715

### DIFF
--- a/s3-uploads.php
+++ b/s3-uploads.php
@@ -4,7 +4,7 @@
 Plugin Name: S3 Uploads
 Description: Store uploads in S3
 Author: Human Made Limited
-Version: 3.0.8
+Version: 3.0.10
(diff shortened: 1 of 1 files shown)


joehoyle (approved):

joehoyle:
Thanks, would like to automate this somehow, but seems tricky as the current process is to create a tag / github release from that tag, so by definition it would be too late to auto-update the file on tagging.

BrookeDot:
Yeah, likely GitHub actions might be able to input the build number as the version. 10up has something similar, for WordPress.org deploys (https://github.com/10up/action-wordpress-plugin-deploy?) but the idea might be worth exploring. 

I opened #715 to discuss this outside of a PR.

PR #722, opened 2025-08-28
Fix autoloading

dingo-d:
Not sure why, the autoloading is skipped.

TBH I'm unsure how this worked, but it errored out to me locally so I'm putting it back in.

Hopefully this solves the issue where plugin stream class file is wrong...


dingo-d:
Sorry for the PR, this was intended for my fork 🙈

PR #732, opened 2026-01-16
feat: directory rename

Luc-cpl:
This pull request significantly enhances the S3 stream wrapper by adding robust support for renaming directories (not just files) within S3 buckets, ensuring correct handling of special characters in object keys (such as spaces and parentheses), and improving test coverage for these scenarios. The update introduces batch operations for efficient directory renames and adds comprehensive tests to verify correct behavior and encoding.

**Major enhancements to S3 stream wrapper directory rename functionality:**

* Added support for renaming directories (not just single files) by recursively copying all objects under the source prefix to the new destination, then deleting the originals, using AWS S3 batch operations for efficiency. This includes handling directory markers and cache invalidation.
* Implemented the `isDirectoryPrefix` helper to accurately detect when a given key represents a directory (prefix) in S3, ensuring the correct rename logic is applied.
* Introduced the `encodeCopySource` method to correctly URL-encode each path segment for S3's `CopySource` parameter, preventing failures when object keys contain spaces or special characters.

**Test improvements and validation:**

* Added extensive tests for directory renaming, including cases with nested files, empty directories, and files with special characters in their names, to ensure the new logic works as intended.
* Included tests to verify that the `encodeCopySource` method is used in the directory rename implementation and that file renames with special characters succeed.

**Dependency update:**

* Added import of `Aws\CommandPool` to enable batch operations for copying objects in directory renames.

### DIFF
--- a/inc/class-stream-wrapper.php
+++ b/inc/class-stream-wrapper.php
@@ -738,13 +738,12 @@ class Stream_Wrapper {
 	}
 
 	/**
-	 * Called in response to rename() to rename a file or directory. Currently
-	 * only supports renaming objects.
+	 * Called in response to rename() to rename a file or directory.
 	 *
-	 * @param string $path_from the path to the file to rename
-	 * @param string $path_to   the new path to the file
+	 * @param string $path_from the path to the file or directory to rename
+	 * @param string $path_to   the new path to the file or directory
 	 *
-	 * @return bool true if file was successfully renamed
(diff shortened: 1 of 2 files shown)


samucsouza (approved):

Luc-cpl:
> I think this largely makes sense. @Luc-cpl did you use AI to write this PR?

Hi @joehoyle.
I used AI only to write the tests.

PR #659, opened 2024-01-02
Update s3 sdk to 3.27

sammyaxe:

PR #740, opened 2026-02-23
Fix test setup for modern Minio compatibility

mikelittle:
## Summary

In preparation for upgrading the old version of `aws-sdk-php` (see [PR 659](https://github.com/humanmade/S3-Uploads/pull/659), I brought the tests back into a working state.
For now, still sticking with Minio. 

- Update `run-tests.sh` to use current Minio env vars (`MINIO_ROOT_USER`/`MINIO_ROOT_PASSWORD`) replacing deprecated `MINIO_ACCESS_KEY`/`MINIO_SECRET_KEY`
- Create the `tests` bucket via `mc` CLI instead of relying on filesystem directories, which no longer works with modern Minio
- Use Docker `--link` for container networking and hostname `minio` instead of hardcoded Docker bridge IP `172.17.0.2`
- Add `use_path_style_endpoint` to S3 client params to avoid virtual hosted-style URLs that Minio can't resolve
- Set `WP_TESTS_PHPUNIT_POLYFILLS_PATH` constant required by the newer WP test suite

## Test plan

- [x] Ran `composer test` locally — all 27 tests pass with 0 errors

### DIFF
--- a/tests/bootstrap.php
+++ b/tests/bootstrap.php
@@ -18,7 +18,8 @@ tests_add_filter( 'muplugins_loaded', '_manually_load_plugin' );
 
 
 tests_add_filter( 's3_uploads_s3_client_params', function ( array $params ) : array {
-	$params['endpoint'] = 'http://172.17.0.2:9000';
+	$params['endpoint'] = 'http://minio:9000';
+	$params['use_path_style_endpoint'] = true;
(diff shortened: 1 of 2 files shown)


wisyhambolu (approved):
Looks good, test successfully locally.

PR #742, opened 2026-03-09
Support default private uploads with Tachyon image resizing

mikelittle:
- Extract presigned URI to access both query string and host
- Add MIME type detection to handle images and non-images differently
- For images: preserve original URL host (CDN/site domain) and pass S3 host as X-Amz-S3-Host parameter so Tachyon can process resizing while maintaining correct signature
- For non-images (PDFs, etc.): use full presigned S3 URL to bypass CloudFront signature validation issues
- Addresses issue https://github.com/humanmade/altis-media/issues/162


mikelittle:
This PR is no longer needed. The feature was implemented without requiring this.

Issue #746, opened 2026-04-07
get_s3_location_for_url() doesn't recognise regional S3 hostnames


mikelittle:
## `get_s3_location_for_url()` doesn't recognise regional S3 hostnames (`bucket.s3.{region}.amazonaws.com`)

`Plugin::get_s3_location_for_url()` (`inc/class-plugin.php` around line 280) only resolves URLs that match one of two patterns:

1. The legacy global S3 form: `https://{bucket}.s3.amazonaws.com/...`
2. The site's `wp_upload_dir()['baseurl']`.

It does **not** recognise the standard **regional** S3 form: `https://{bucket}.s3.{region}.amazonaws.com/...`. As a result, any URL that arrives in this form passes through `add_s3_signed_params_to_attachment_url()` unchanged (because `get_s3_location_for_url()` returns `null`), and the file ends up referenced by an unsigned URL — broken for any private attachment.

### How we hit it

In an environment configured with `S3_UPLOADS_REGION=eu-west-1` (or similar), `wp_get_attachment_url($id)` for a private PDF returns a presigned URL like:

```
https://hmn-uploads-eu.s3.eu-west-1.amazonaws.com/uploads/2026/04/my-test-ebook.pdf?X-Amz-Signature=...
```

WordPress core's `image_downsize()` then derives the cover-image sub-size URL by taking `dirname()` of that URL and substituting the cover JPEG filename, dropping the query string in the process:

```
https://hmn-uploads-eu.s3.eu-west-1.amazonaws.com/uploads/2026/04/my-test-ebook-pdf-106x150.jpg
```

This unsigned URL is then passed through the `wp_get_attachment_image_src` filter chain. S3 Uploads' `add_s3_signed_params_to_attachment_image_src()` calls `add_s3_signed_params_to_attachment_url()` → `get_s3_location_for_url()` → returns `null` because neither prefix check matches → URL is returned unsigned → browser hits 403 → media library tile renders broken.

We worked around this in our private-media module by rewriting the URL host to the upload baseurl before calling `add_s3_signed_params_to_attachment_url()`, but that workaround shouldn't be necessary — `get_s3_location_for_url()` should resolve regional S3 URLs natively, since the regional form is the AWS-recommended default and what the AWS SDK signs against.

### Suggested fix

Make the first check in `get_s3_location_for_url()` match both forms — something like:

```php
// Match both legacy and regional S3 hostnames:
//   https://{bucket}.s3.amazonaws.com/...
[...]

PR #748, opened 2026-06-08
Fix get_s3_location_for_url() to match regional S3 hostnames

thisismyurl:
## Problem

\`get_s3_location_for_url()\` only matches the legacy global S3 hostname:

\`\`\`php
\$s3_url = 'https://' . \$this->get_s3_bucket() . '.s3.amazonaws.com/';
if ( strpos( \$url, \$s3_url ) === 0 ) { ... }
\`\`\`

AWS's current recommended hostname form for any configured region is \`bucket.s3.{region}.amazonaws.com\`. When S3 Uploads is configured with a non-\`us-east-1\` region, the AWS SDK generates signed private attachment URLs using the regional hostname. The method returns \`null\` for those URLs, so \`add_s3_signed_params_to_attachment_url()\` passes them through unsigned — breaking private attachment access on regionally-configured installs.

This is the issue described in #746.

## Fix

Replace the \`strpos()\` prefix check with a regex that matches both the legacy and regional hostname forms:

\`\`\`php
\$s3_pattern = '#^https?://' . preg_quote( \$this->get_s3_bucket(), '#' ) . '\.s3(?:\.[a-z0-9-]+)?\.amazonaws\.com/#i';
if ( preg_match( \$s3_pattern, \$url ) ) { ... }
\`\`\`

The optional \`(?:\.[a-z0-9-]+)?\` group matches the region segment when present and is absent for the legacy global endpoint — so both \`bucket.s3.amazonaws.com\` and \`bucket.s3.us-west-2.amazonaws.com\` work correctly.

## Test

Added \`test_get_s3_location_for_url_regional_hostname()\` asserting that a \`us-west-2\` regional URL resolves to the correct bucket and key.

---

*Developed with AI assistance (GitHub Copilot + Claude) for code review; the fix, test, and description are my own.*

### DIFF
--- a/inc/class-plugin.php
+++ b/inc/class-plugin.php
@@ -278,8 +278,9 @@ class Plugin {
 	 * @return array{bucket: string, key: string, query: string|null}|null
 	 */
 	public function get_s3_location_for_url( string $url ) : ?array {
-		$s3_url = 'https://' . $this->get_s3_bucket() . '.s3.amazonaws.com/';
-		if ( strpos( $url, $s3_url ) === 0 ) {
+		// Match both legacy (bucket.s3.amazonaws.com) and regional (bucket.s3.{region}.amazonaws.com) hostnames.
+		$s3_pattern = '#^https?://' . preg_quote( $this->get_s3_bucket(), '#' ) . '\.s3(?:\.[a-z0-9-]+)?\.amazonaws\.com/#i';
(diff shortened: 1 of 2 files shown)


cavaquinho93:
Is there any likelihood of this PR being reviewed soon? It's quite high priority IMO as currently signed parameters are not working for image sizes other than the original.

PR #751, opened 2026-07-07
Return filesize from Image_Editor_Imagick::_save()

### LINKED ISSUE #750: `Image_Editor_Imagick::_save()` drops `filesize` from its return value — attachment meta ends up with `filesize => 0` after edits in the media library image editor
## Summary

Since WP 6.0, `WP_Image_Editor::_save()` includes a `filesize` key in its return array ([core ticket #49412](https://core.trac.wordpress.org/ticket/49412), see [`WP_Image_Editor_Imagick::_save()`](https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/class-wp-image-editor-imagick.php)).

`S3_Uploads\Image_Editor_Imagick::_save()` overrides that method (to save to a local temp file and copy it to S3) and builds its own response array — which omits `filesize`:

https://github.com/humanmade/S3-Uploads/blob/master/inc/class-image-editor-imagick.php#L124-L131

The `filesize` from `parent::_save()` (which is correct — the temp file is byte-identical to what gets copied to S3) is simply discarded.

## Impact

1. **Media library image editor (crop/rotate/scale):** `wp_save_image()` in `wp-admin/includes/image-edit.php` does `$meta['filesize'] = $saved_image['filesize'];` — with this plugin active that's a PHP "Undefined array key" warning, and the attachment meta is persisted without a usable `filesize`. Subsequent operations that read it (`wp_prepare_attachments_for_js()`, `wp_restore_image()`'s `backup_sizes` writes) either warn again or fall back to a remote `filesize()` stat against S3 on every call — exactly the slow HEAD requests that #492 / #493 were about.
2. **Sub-sizes on every upload:** `make_subsize()` / `multi_resize()` return this same `_save()` array, so `sizes[*]['filesize']` is missing from attachment meta for all thumbnails, always.

## Why #493 doesn't cover this

`set_filesize_in_attachment_meta()` (added in #493) is hooked on `wp_generate_attachment_metadata`, which only fires when metadata is generated (upload, sideload, regenerate). The image editor's save path (`wp_save_image()`) assembles `$meta` by hand and persists it via `wp_update_attachment_metadata()` directly, so the backfill never runs there. It also only sets the top-level `filesize`, not the per-size ones.

## Reproduction

1. WP ≥ 6.0 with S3-Uploads active.
2. Upload a JPEG, open it in the media library image editor, rotate or crop, save.
3. Inspect `wp_get_attachment_metadata( $id )` — `filesize` is missing (and a PHP warning was logged from `image-edit.php`). Per-size `filesize` values are missing even on step 2's initial upload.

## Suggested fix

Return the `filesize` that `parent::_save()` already computed:

```php
$response = [
    'path'      => $filename,
    'file'      => wp_basename( apply_filters( 'image_make_intermediate_size', $filename ) ),
    'width'     => $this->size['width'],
    'height'    => $this->size['height'],
    'mime-type' => $mime_type,
    'filesize'  => $save['filesize'] ?? wp_filesize( $filename ),
];
```

Using the parent's value avoids a remote stat; the `wp_filesize()` fallback only applies if a future WP version stops returning it. This also makes the #493 backfill effectively redundant (though it's harmless to keep, being guarded by `! isset`).

karlisb:
Fixes #750.

`Image_Editor_Imagick::_save()` rebuilds its response array after copying the temp file to S3, but omits the `filesize` key that `WP_Image_Editor::_save()` has returned since WP 6.0 ([core ticket #49412](https://core.trac.wordpress.org/ticket/49412)).

As a result:

- saves from the media library image editor (`wp_save_image()`) trigger an "Undefined array key" PHP warning and persist attachment meta without a usable `filesize`;
- `sizes[*]['filesize']` is missing from attachment meta for all sub-sizes on every upload;
- code paths that read the value fall back to remote `filesize()` stats against S3 — the slow HEAD requests #492 / #493 aimed to eliminate (the #493 backfill hooks `wp_generate_attachment_metadata`, which never fires on the image-editor save path).

This change returns the `filesize` that `parent::_save()` already computed for the local temp file — byte-identical to the object copied to S3, so no extra remote request is made. The `wp_filesize()` fallback only applies if a future WP version stops returning the key. Docblocks updated to match.

### DIFF
--- a/inc/class-image-editor-imagick.php
+++ b/inc/class-image-editor-imagick.php
@@ -78,7 +78,7 @@ class Image_Editor_Imagick extends WP_Image_Editor_Imagick {
 	 * @param Imagick $image
 	 * @param ?string $filename
 	 * @param ?string $mime_type
-	 * @return WP_Error|array{path: string, file: string, width: int, height: int, mime-type: string}
+	 * @return WP_Error|array{path: string, file: string, width: int, height: int, mime-type: string, filesize: int}
 	 */
 	protected function _save( $image, $filename = null, $mime_type = null ) {
 		list( $filename, $extension, $mime_type ) = $this->get_output_format( $filename, $mime_type );
@@ -96,7 +96,7 @@ class Image_Editor_Imagick extends WP_Image_Editor_Imagick {
 		}
 
 		/**
-		 * @var WP_Error|array{path: string, file: string, width: int, height: int, mime-type: string}
+		 * @var WP_Error|array{path: string, file: string, width: int, height: int, mime-type: string, filesize: int}
 		 */
 		$parent_call = parent::_save( $image, $temp_filename !== false ? $temp_filename : $filename, $mime_type );
 
@@ -127,6 +127,7 @@ class Image_Editor_Imagick extends WP_Image_Editor_Imagick {
 			'width'     => $this->size['width'] ?? 0,
 			'height'    => $this->size['height'] ?? 0,
 			'mime-type' => $mime_type,
+			'filesize'  => $save['filesize'] ?? wp_filesize( $filename ),
 		];
 
 		return $response;


inc/class-image-editor-imagick.php
```diff
@@ -127,6 +127,7 @@ protected function _save( $image, $filename = null, $mime_type = null ) {
 			'width'     => $this->size['width'] ?? 0,
 			'height'    => $this->size['height'] ?? 0,
 			'mime-type' => $mime_type,
+			'filesize'  => $save['filesize'] ?? wp_filesize( $filename ),
```
joehoyle, line 130: Would it not be better to get the filesize before it's uploaded to s3? That would reduce / remove a network call.

joehoyle (changes requested):

inc/class-image-editor-imagick.php
karlisb: It already does, as `$save` is a reference the return value of `parent::save()`.
You're right that `wp_filesize` is a fallback that will make an extra `HEAD` request but only so this wouldn't break if by any reason Wordpress would revert the introduced filesize in 6.0. In practice this shouldn't be executed ever.
Would you rather me implement a fallback that stats the temp file (a few lines up) before discarding it rather that risking an extra `HEAD` request if Wordpress goes berserk? 😄

